PR #8, opened 2017-10-12
Configuration for new Dashboard functionality

kloostert:
This should enable the new unit-test-grouping-by-endpoint functionality of the Dashboard to be used in the Zeeguu-API Travis integration.

The `dashboard_test.cfg` file in this PR does not contain the SUBMIT_RESULTS_URL option, which means that the test results will not be uploaded to the live deployment of the Dashboard. 
This way, unwanted Travis test results will not be displayed on the live Dashboard.

If the results should be uploaded to the live Dashboard again, simply adding the following line to the config file will do the trick:
`SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results`

### DIFF
--- a/dashboard_test.cfg
+++ b/dashboard_test.cfg
@@ -1,5 +1,4 @@
 [dashboard]
 TEST_DIR=./zeeguu_api/tests/
+LOG_DIR=./
 N=5
-SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results
-
--- a/dashboard_test.cfg
+++ b/dashboard_test.cfg
@@ -2,3 +2,4 @@
 TEST_DIR=./zeeguu_api/tests/
 LOG_DIR=./
 N=5
+SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ setuptools.setup(
     dependency_links=[
             "git+https://github.com/mircealungu/zeeguu-core.git#egg=zeeguu",
             "git+https://github.com/mircealungu/python-translators.git#egg=python_translators",
+            "https://github.com/flask-dashboard/Flask-Monitoring-Dashboard/tarball/development#egg=flask_monitoring_dashboard",
         ],
     install_requires=("flask>=0.10.1",
                       "Flask-SQLAlchemy",
--- a/setup.py
+++ b/setup.py
@@ -14,8 +14,8 @@ setuptools.setup(
     description="API for Zeeguu, a project that aims to accelerate vocabulary acquisition in a second language",
     keywords=" API, second language acquisition",
     dependency_links=[
-            "git+https://github.com/mircealungu/zeeguu-core.git#egg=zeeguu",
-            "git+https://github.com/mircealungu/python-translators.git#egg=python_translators",
+            "https://github.com/zeeguu-ecosystem/Zeeguu-Core/tarball/master#egg=zeeguu",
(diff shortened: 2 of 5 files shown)


dashboard_test.cfg
```diff
@@ -1,5 +1,4 @@
 [dashboard]
 TEST_DIR=./zeeguu_api/tests/
+LOG_DIR=./
 N=5
-SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results
-
```
mircealungu: i'm happy that this is not hardcoded anymore :) 
but how come is it not needed?

setup.py
```diff
@@ -16,6 +16,7 @@
     dependency_links=[
             "git+https://github.com/mircealungu/zeeguu-core.git#egg=zeeguu",
```
mircealungu: can update this too to: https://github.com/zeeguu-ecosystem/Zeeguu-Core

setup.py
```diff
@@ -16,6 +16,7 @@
...
             "git+https://github.com/mircealungu/python-translators.git#egg=python_translators",
```
mircealungu: and this to: 

   https://github.com/zeeguu-ecosystem/Python-Translators

zeeguu_api/app.py
```diff
@@ -25,7 +25,7 @@
 
 try:
     import dashboard
-    dashboard.config.from_file('/home/mircea/zee/http/api/dashboard.cfg')
+    dashboard.config.from_file('/home/travis/build/zeeguu-ecosystem/Zeeguu-API/dashboard_test.cfg')
```
mircealungu: hmm. instead of fighting for this string, maybe it's good to use an envvar for the location of the dashboard.cfg? e.g. DASHBOARD_CFG ?

dashboard_test.cfg
kloostert: It's still needed, but for testing purposes, I turned the uploading of unit test results off.
This way, the live deployment of the Dashboard would not receive many Travis test results for the same Zeeguu-API version while I was testing the Travis configuration. 
But this can easily be turned back on again by adding the submit-results-url to the config file whenever necessary.

zeeguu_api/app.py
kloostert: Yes, this would be better. I can implement this such that the Dashboard will take the envvar value when specified, and otherwise takes the dashboard.config.from_file().

zeeguu_api/app.py
```diff
@@ -25,7 +25,7 @@
 
 try:
     import dashboard
-    dashboard.config.from_file('/home/travis/build/zeeguu-ecosystem/Zeeguu-API/dashboard_test.cfg')
+    dashboard.config.from_file('/home/mircea/zee/http/api/dashboard.cfg')
```
mircealungu: @kloostert -- this seems to still be hardcoded, and not using the envvar, right ? :)

kloostert: Right, I reverted that line back to what it was before. 
But specifying envvar `DASHBOARD_CONFIG` will override this location string.
Thus, whenever the envvar is specified, it will use that one, else the location string provided in the from_file() call.

mircealungu: coool! but if we use the envvar then i think we should drop completely this hardcoded string... its's security paranoia... but, we are giving away info about the state of our server :)

mircealungu: Also, a note should be added here about the envvar: https://github.com/flask-dashboard/Flask-Monitoring-Dashboard

kloostert: Ok, added some info about it in the setup section of the readme (on the dev branch).
The line which configures the dashboard can be set to `dashboard.config.from_file(None)`.
This will take the envvar instead of the location string.

PR #9, opened 2017-10-19
Improvements on unit test result grouping

kloostert:
This configuration should allow for the grouping of unit test results based on the endpoint that they test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ cache:
 install: "python setup.py develop"
 
 script: 
-  - "export DASHBOARD_CONFIG=/home/travis/build/kloostert/Zeeguu-API/dashboard_test.cfg"
-  - "export DASHBOARD_LOG_DIR=/home/travis/build/kloostert/Zeeguu-API/"
+  - "export DASHBOARD_CONFIG=/home/travis/build/zeeguu-ecosystem/Zeeguu-API/dashboard_test.cfg"
+  - "export DASHBOARD_LOG_DIR=/home/travis/build/zeeguu-ecosystem/Zeeguu-API/"
   - "./run_tests.sh"
   - "python -m dashboard.collect_performance"
--- a/dashboard_test.cfg
+++ b/dashboard_test.cfg
@@ -1,4 +1,4 @@
 [dashboard]
 TEST_DIR=./zeeguu_api/tests/
 N=5
-SUBMIT_RESULTS_URL=http://143.176.113.105:5000/dashboard/submit-test-results
+SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results


dashboard_test.cfg
```diff
@@ -1,5 +1,4 @@
 [dashboard]
 TEST_DIR=./zeeguu_api/tests/
-LOG_DIR=./
 N=5
-SUBMIT_RESULTS_URL=https://zeeguu.unibe.ch/api/dashboard/submit-test-results
+SUBMIT_RESULTS_URL=http://143.176.113.105:5000/dashboard/submit-test-results
```
mircealungu: @kloostert -- is this correct?

.travis.yml
```diff
@@ -21,6 +21,7 @@ cache:
 install: "python setup.py develop"
 
 script: 
+  - "export DASHBOARD_CONFIG=/home/travis/build/kloostert/Zeeguu-API/dashboard_test.cfg"
+  - "export DASHBOARD_LOG_DIR=/home/travis/build/kloostert/Zeeguu-API/"
   - "./run_tests.sh"
-  - "export DASHBOARD_CONFIG=/home/travis/build/zeeguu-ecosystem/Zeeguu-API/dashboard_test.cfg"
-  - "python collect_performance.py"
+  - "python -m dashboard.collect_performance"
```
mircealungu: cool that you've moved the script in dashboard :) 👍

.travis.yml
```diff
@@ -21,6 +21,7 @@ cache:
 install: "python setup.py develop"
 
 script: 
+  - "export DASHBOARD_CONFIG=/home/travis/build/kloostert/Zeeguu-API/dashboard_test.cfg"
+  - "export DASHBOARD_LOG_DIR=/home/travis/build/kloostert/Zeeguu-API/"
```
mircealungu: privacy issue? :)

zeeguu_api/app.py
```diff
@@ -25,7 +25,7 @@
 
 try:
     import dashboard
-    dashboard.config.from_file('/home/mircea/zee/http/api/dashboard.cfg')
+    dashboard.config.from_file(None)
```
mircealungu: this is misleading... 
maybe implement `from` in such a way that you can call it either 
    dash.config.from(file='/home/thijs/...')
or 
   dash.config.from(envvar='DASHBOARD_CONFIG')

that would be so much more readable :)

zeeguu_api/api/__init__.py
```diff
@@ -14,4 +14,19 @@
...
+import datetime
+import os
+log_dir = os.getenv('DASHBOARD_LOG_DIR')
+@api.after_request
+def after_request(response):
+    t1 = str(datetime.datetime.now())
+    log = open(log_dir + "endpoint_hits.log", "a")
+    log.write("\"{}\",\"{}\"\n".format(t1, request.endpoint))
+    log.close()
+    return response
```
mircealungu: you should be able to put this code in the dashboard/__init__.py#bind(app) function... you only seem to need the api variable and that is available at the point where the api calls dashboard.bind(app=app)... so you could simply call dashboard.bind(app=app, blueprints=[api]) and in that way you can monitor even multiple blueprints if needed. 

see what i mean?

.travis.yml
kloostert: I'll look into making the paths relative instead of absolute, would that fix the privacy issue?

dashboard_test.cfg
kloostert: No this is the IP of a test server on which I deployed the dashboard for testing it out a bit, but the second commit in this PR actually changes the url back to the Zeeguu dashboard.

zeeguu_api/app.py
kloostert: Yes, I can definitely do this.

zeeguu_api/api/__init__.py
kloostert: I see, I will try this out.

mircealungu: @kloostert  - Thijs, what's happening with this PR?

PR #13, opened 2018-03-19
Feature/api transition

LarsHoldijk:
This pull request contains the code for resetting a password and a new method for returning an error code and a message to the clients. Previously used methods for aborting in the API such as `flask.abort()` will still work.

Note that this pull request goes hand in hand with the change to the core found in the latest pull request. As this change in the core is not yet merged the pull request will not pass travis.


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,3 @@
+{
+  "tests_zeeguu_api/test_session.py::SessionTests::test_reset_password_unkown_email": true
```
mircealungu: this file seems to have been introduced by your PR. 
correct?

default_api.cfg
```diff
@@ -8,3 +8,9 @@ DEBUG=True
 HOST="0.0.0.0"
 API_PORT=9001
 SECRET_KEY="lalalal"
+
+## The SMTP config information below is used for
+## sending password reset requests.
+SMTP_SERVER = ''
+SMTP_USERNAME= ''
+SMTP_PASSWORD= ''
```
mircealungu: nice! forgot about this!

tests_zeeguu_api/test_session.py
```diff
@@ -18,6 +21,87 @@ def test_create_user(self):
         rv = self.api_post('/add_user/i@i.la',form_data)
         assert len(rv.data) > 1
 
+    def test_create_user_returns_400(self):
```
mircealungu: test_create_user_returns_400_if_password_not_present

this should be the name, right?

tests_zeeguu_api/test_session.py
```diff
@@ -18,6 +21,87 @@ def test_create_user(self):
...
+
+        form_data = dict(
+            code=code,
+            password="updated"
+        )
+        rv = self.api_post('/reset_password/' + TEST_EMAIL, form_data)
+        assert rv.status_code == 200
+
+    def test_reset_password_invalid_code(self):
+        code = UniqueCode(TEST_EMAIL)
```
mircealungu: nice and clean test.

tests_zeeguu_api/test_session.py
```diff
@@ -18,6 +21,87 @@ def test_create_user(self):
...
+
+        form_data = dict(
+            password="updated"
+        )
+
+        rv = self.api_post('/session/' + TEST_EMAIL, form_data)
+
+        assert rv.status_code == 200
+
+    def test_reset_password_cant_use_old_password(self):
```
mircealungu: nice tests here @LarsHoldijk !

zeeguu_api/api/sessions.py
```diff
@@ -20,14 +23,14 @@ def add_user(email):
     password = request.form.get("password")
     username = request.form.get("username")
     if password is None:
-        flask.abort(400)
+        return make_error(400, "Invalid value")
```
mircealungu: should this not be something more precise?

zeeguu_api/api/sessions.py
```diff
@@ -108,6 +109,51 @@ def get_anon_session(uuid):
...
+@api.route("/send_code/<email>", methods=["POST"])
+@cross_domain
+def send_code(email):
+    """
+    This endpoint generates a unique code that will be used to allow
+    the user to change his/her password. The unique code is send to
+    the specified email address.
+    """
+    code = UniqueCode(email)
+    zeeguu.db.session.add(code)
```
mircealungu: i know that most of the API is written like this, 
but I think it's nicer and cleaner if at the top
of the file we simply declare session = zeeguu.db.session
and then we write code like this simply 

    session.add(code)
    session.commit()

zeeguu_api/api/utils/abort_handling.py
```diff
@@ -0,0 +1,9 @@
+from flask import jsonify
+
+
+def make_error(status_code, message,):
+    response = jsonify({
+        'message': message,
+    })
+    response.status_code = status_code
+    return response
```
mircealungu: always newline after last char in file.

zeeguu_api/api/sessions.py
LarsHoldijk: Totally agree. I used db_session instead to not get confused with the flask session. 

I defined db_session in the init file and updated the other endpoints accordingly.

zeeguu_api/api/sessions.py
LarsHoldijk: I added a test for password length here now. I noticed it was still missing. The message now reads "Password should be at least 4 characters long" if no password is given or too short.

zeeguu_api/api/utils/abort_handling.py
LarsHoldijk: Is there any specific reason for doing this? I (normally) always do this but I never stopped a think why actually.

mircealungu: newline is the official line terminator, and it's nice to terminate all the lines, even the last :)

zeeguu_api/api/sessions.py
mircealungu: nice idea with the db_session. 
i just realized yesterday that there's so many things called session in the system, that i'm sorry for the dev-team. flask.session where we save the session_id which is the  session column of the Session table... :)

PR #17, opened 2018-04-07
Added user_working_session methods

carlospaz7:


zeeguu_api/api/__init__.py
```diff
@@ -19,4 +19,5 @@
 from . import user_statistics
 from . import recommendations
 from . import user_article
-from . import user_articles
\ No newline at end of file
+from . import user_articles
+from . import user_working_session
```
mircealungu: the last line in the file also deserves a newline terminator ;)

zeeguu_api/api/user_working_session.py
```diff
@@ -0,0 +1,173 @@
...
+@with_session
+def working_sessions_by_user():
+    """
+        Returns user_working_sessions data based on the provided filters for the specified user
+
+        Parameters:
+            user_id
+            from_date   (using ISO-8601 format)
+            to_date     (using ISO-8601 format)
+            is_active
```
mircealungu: beautiful docstring @carlospaz7 !!!

zeeguu_api/api/user_working_session.py
```diff
@@ -0,0 +1,173 @@
...
+            [
+                147,
+                "2018-02-24T09:11:11",
+                null,
+                61
+            ],
+            ...
+        ]
+    """
+    user_id =  int(flask.request.args['user_id'])
```
mircealungu: i wonder whether i'm missing something.

but if not, then the user_id does not have to be passed explicitly in the call. 
the user id can be extracted from the session id, because the session id is linked to a user id.

or maybe is this function thought to be used by a teacher to obtain info about a student?
in that case, we must add a check to make sure that the one who requests info is teacher for the cohort of the one about which the info is being requested. however, the dashboard-dev-team has already created functionality for this, so we should reuse it.

zeeguu_api/api/user_working_session.py
```diff
@@ -0,0 +1,173 @@
...
+                "2017-05-19T14:58:58",
+                63399,
+                0
+            ],
+            ...
+        ]
+    """
+    cohort_id =  int(flask.request.args['cohort_id'])
+    from_date = flask.request.args['from_date']
+    to_date = flask.request.args['to_date']
```
mircealungu: ditto. both
- very elegant
- must pay attention with accessibility restrictions

zeeguu_api/api/user_working_session.py
```diff
@@ -0,0 +1,173 @@
...
+    is_active = bool(flask.request.args['is_active'])
+    active_sessions = UserWorkingSession.find_by_cohort(cohort_id, from_date, to_date, is_active)
+    result = [[item.id, item.user_id, item.start_time.isoformat(), item.article_id, item.duration] for item in active_sessions]
```
mircealungu: actually, both here and in the previous example we could have returned a dictionary instead of an array for increased readability of the client-side code. if we return a dictionary the client can write code like: `result['id']`, `result['article_id']` as opposed to `result[0]` and `result[2]`. 

what do you think?

tests_zeeguu_api/test_working_session.py
```diff
@@ -0,0 +1,53 @@
...
+        working_session = WorkingSessionRule().w_session
+        self.user_id = working_session.user_id
+        self.article_id = working_session.article_id
+        self.is_active = working_session.is_active
+        self.cohort_id = working_session.user.cohort_id
+
+    def test_working_sessions_by_user(self):
+        parameters = dict(
+            user_id = self.user_id,
+            from_date = (datetime.now() - timedelta(minutes=randint(75000, 75000))).isoformat(),
```
mircealungu: as far as I can tell `randint(75000, 75000)` is always 75000.

mircealungu: @carlospaz7  - is this PR still active?
Then why no answer to this comment?
If not, let's close it... it's getting old :)

carlospaz7: In the end the development team used their own API and got the data directly from the CORE, also many things have changed since this version (in this version of the API I still called it working session, instead of reading and exercise) I think we can close the PR and only if is needed, I can create the APIs for the reading and exercise sessions.
I totally forgot about it.

mircealungu:
Thanks Carlos!
I agree with your analysis.
Close it please then.

M.

carlospaz7:
A direct call to the CORE from the teacher dashboard API, therefore this pull request is no longer needed.

PR #20, opened 2018-04-17
Dashboard

Oli26:


zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+import zeeguu
+from zeeguu.model import User, Cohort, UserActivityData, Session, Language
+import sqlalchemy
+from flask import jsonify
+import json
+import random
+import datetime
+
+# class function wrapper
+def class_function_checker(class_id):
```
mircealungu: comment is misleading function name does not help much.
- this is not a wrapper
- comments should say *why* a function is needed 
- function name it should say *what* it does ... in our case it seems to imply that it checks a class... checks for what?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+    for l in link:
+        if l.user_id == flask.g.user.id:
+            return True
+    return False
+
+
+# Takes user_id and returns user.name that corresponds
+#@api.route("/get_user_name/<id>", methods=["GET"])
+@with_session
+def get_user_name(id):
```
mircealungu: this seems to be weak design.
seems to imply that if i'm logged in, 
and thus have a session, i can find out the name/id 
matching of everybody else in the system. 

i guess there should be a check whether the logged
in user is a teacher for the user for which we ask info 
about, right?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+    user = User.query.filter_by(id=id).one()
+    return user.name
+
+
+# Checks to see if user has a valid active session
+@api.route("/has_session", methods=["GET"])
```
mircealungu: this is not used anywhere. 
why is it needed? 

seems to replicate the behavior in @with_session which can be triggered by the /validate endpoint (in zeeguu_api/api/sessions)

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+@api.route("/get_class_permissions/<id>", methods=["GET"])
+@with_session
+def get_class_permissions(id):
+    if(class_function_checker(id)):
+        return jsonify(1)
+    return jsonify(0)
+
+@api.route("/get_user_permissions/<id>", methods=["GET"])
+@with_session
+def get_user_permissions(id):
```
mircealungu: method docstring would be nice here too.

zeeguu_api/api/dashboard.py
mircealungu: also, seems to be superseded by /get_user_info below, right?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+
+# Asking for a nonexistant cohort will cause .one() to crash!
+# Takes cohort_id and returns all users belonging to that cohort
+@api.route("/get_users_from_class/<id>", methods=["GET"])
+@with_session
+def get_users_from_class(id):
+    if(not class_function_checker(id)):
+        flask.abort(401)
+    c = Cohort.query.filter_by(id=id).one()
+    if not c is None:
```
mircealungu: ca can not be None. 
either it returns something, or throws exception.
thus there should be a try:/exxcept: here.
or?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+@api.route("/get_user_info/<id>", methods=["GET"])
+@with_session
+def wrapper_to_json_user(id):
+    if(not get_user_permissions(id)):
+        flask.abort(401)
+    return jsonify(get_user_info(id))
+
+# Gets user words info
+@with_session
+def get_user_info(id):
```
mircealungu: if you annotate the routes with @with_session you don't have to 
annotate the functions that are called from them anymore since
they could have not been called if there was no session. make
sure that this does not happen to other helper functions. 

also, for such a "helper" function, which is not supposed to be used
elsewhere than this file, you could consider using the _get_user_info
naming convention.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+            'last_article': 'place holder article'
+        }
+        return dictionary
+    except ValueError:
+        flask.abort(400)
+
+
+
+
+# Removes class
```
mircealungu: comment repeats the method name. 
can you be more informative? 

also, i'm not sure it makes anything easier if
we talk about classes and cohorts in parallel.

also, a note should be made that only the association
between teacher and cohort is deleted. the students
are still left in the cohort, right? 

ah, now i see: this works only if there are no students.
well, could be added as a comment maybe..

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+# Takes cohort_id and reuturns dictionary with relevant class variables
+@api.route("/get_class_info/<id>", methods=["GET"])
+@with_session
+def wrapper_to_json_class(id):
+    if(not class_function_checker(id)):
+        flask.abort(401)
+    return jsonify(get_class_info(id))
+
+
+def get_class_info(id):
```
mircealungu: rename to _get_class_info

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+    max_students = c.max_students
+    cur_students = c.cur_students
+    class_language_id = c.class_language_id
+    class_language = Language.query.filter_by(id=class_language_id).one()
+    d = {'id': str(id), 'class_name': class_name, 'inv_code': inv_code, 'max_students': max_students,
+         'cur_students': cur_students, 'class_language_name': class_language.name, 'class_id': id}
+    return d
+
+
+# Takes two inputs (user_id, cohort_id) and links them other in teacher_cohort_map table.
```
mircealungu: links them other?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+# url input in format <user_id>/<cohort_id>
```
mircealungu: url input format... this is visible in the route def below, or do you mean something else?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+#@api.route("/link_teacher_class/<user_id>/<cohort_id>", methods=["POST"])
+def link_teacher_class(user_id, cohort_id):
+    from zeeguu.model import TeacherCohortMap
+    user = User.find_by_id(user_id)
+    cohort = Cohort.find(cohort_id)
+    zeeguu.db.session.add(TeacherCohortMap(user, cohort))
```
mircealungu: you've already imported db in the scope, so you can say db.sesison(...)

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+# creates a class in the data base. Requires form input (inv_code, class_name, class_language_id, max_students, teacher_id)
+@api.route("/add_class", methods=["POST"])
+@with_session
+def add_class():
+    inv_code = request.form.get("inv_code")
+    class_name = request.form.get("class_name")
+    class_language_id = request.form.get("class_language_id")
+    if len(class_language_id) != 2:
```
mircealungu: ouch... 
i've burned myself with this in the past. 
language codes are not always 2 characters. 
if you look at core.model.Language you'll see that there's 
a situation where it's longer. 

also, to avoid confusion it would be nicer to call this language_code
since in the Language model the id is the primary key and the code
is the iso code.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+        flask.abort(400)
+    class_language = Language.find_or_create(class_language_id)
+    teacher_id = flask.g.user.id
+    max_students = request.form.get("max_students")
+    if int(max_students) < 1:
+        flask.abort(400)
```
mircealungu: we'll need to provide human readable error messages.
for now it's ok but i've added this to the backlog.

https://github.com/RUGSoftEng/2018-ZeeguuDashboard/projects/1#card-9064053

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+    username = request.form.get("username")
+    inv_code = request.form.get("inv_code")
+
+    cohort_id = Cohort.get_id(inv_code)
+    cohort = Cohort.find(cohort_id)
+
+    if not cohort is None:
+        if not password is None:
+            # Checks to see if there is space, if there is it increments the cur_students variables.
+            # Therefor it is essential that you ensure you add the student if this function returns true.
```
mircealungu: therefore :)

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,288 @@
...
+                    flask.abort(400)
+                except sqlalchemy.exc.IntegrityError:
+                    cohort.undo_join()
+                    flask.abort(400)
+            return 'no more space in class!'
+    return 'failed :('
+
+
+#Get user bookmarks
+@api.route("/get_user_stats/<id>", methods=["GET", "POST"])
```
mircealungu: why is this named get user stats if it gets bookmarks?

zeeguu_api/api/dashboard.py
Oli26: Made the comment more clear about function, no longer calling it a wrapper. New function name is 
'has_permission_for_class(class_id)'

zeeguu_api/api/dashboard.py
mircealungu: Oh. I see. Cool! 
In that case maybe the underscore for private functions in the module would help!

zeeguu_api/api/dashboard.py
mircealungu: but can't you use /validate instead?

zeeguu_api/api/dashboard.py
Oli26: Instead of c is None, now use try and catch the exception.

zeeguu_api/api/dashboard.py
Oli26: Replaced all references of zeeguu.db to just db. Defining db as zeeguu.db at the top of the file.

zeeguu_api/api/dashboard.py
Oli26: Fixed :D

zeeguu_api/api/dashboard.py
Oli26: Nevermind, I had already changed get_user_info to make this function redundant. It is now removed!

zeeguu_api/api/dashboard.py
Oli26: Renamed all auxiliary/helper function to have _ convention. Removed all duplicated @with_sessions

zeeguu_api/api/dashboard.py
Oli26: Done

zeeguu_api/api/dashboard.py
Oli26: Removed both the route and this comment. This function is a private auxiliary function for add_class

zeeguu_api/api/dashboard.py
Oli26: Renamed to get_user_bookmarks

zeeguu_api/api/dashboard.py
Oli26: Added more info above remove_class function
Talking about cohorts and classes in parallel is pretty ingrained into the system now :(

zeeguu_api/api/dashboard.py
Oli26: Changed the way in which it checks the language_id!

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+
+    """
+    if (_has_permission_for_class(id)):
+        return jsonify(1)
+    return jsonify(0)
+
+
+@api.route("/get_user_permissions/<id>", methods=["GET"])
+@with_session
+def get_user_permissions(id):
```
mircealungu: the comment is great, but the endpoint name is very vague. 
can we move some of the clarity of the comment i the endpoint name? 

e.g. something like: `/permissions_to_access_user_data`

would be much clearer.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+    try:
+        user = User.query.filter_by(id=id).one()
+        return get_class_permissions(user.cohort_id)
+    except:
+        flask.abort(401)
+
+
+# Asking for a nonexistant cohort will cause .one() to crash!
+# Takes cohort_id and returns all users belonging to that cohort
+@api.route("/get_users_from_class/<id>", methods=["GET"])
```
mircealungu: again, inspired from the comment, the more explanatory name would be `all users in cohort`. agree?
also, given that the HTTP verb is GET there's no need to put get in the name. 
this moves the code a little step in the RESTful direction.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+            info = _get_user_info(u.id)
+            users_info.append(info)
+        return json.dumps(users_info)
+
+    except:
+        flask.abort(400)
+
+
+# Takes cohort_id and reuturns dictionary with relevant class variables
+@api.route("/get_user_info/<id>", methods=["GET"])
```
mircealungu: i'm puzzled here. 
the comment contradicts the code.
why does it say takes cohort_id?

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+            'exercises_done': random.randint(1, 100),
+            'last_article': 'place holder article'
+        }
+        return dictionary
+    except ValueError:
+        flask.abort(400)
+
+
+# Removes class.Can only be called successfuly if the class is empty. Otherwise we return 400 (Value error).
+@api.route("/remove_class/<class_id>", methods=["POST"])
```
mircealungu: ok, sorry, if i was undecided in the previous feedback round.
let's stop talking about classes, and use only `cohort` instead. 
if everywhere else in the project we talk about cohorts, it is 
unprofessional to talk only here about 'classes'. 

so please, let's refactor this to cohort.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+        db.session.commit()
+        return 'removed'
+    except ValueError:
+        flask.abort(400)
+
+
+# Takes Teacher id as input and outputs list of all cohort_ids that teacher owns
+@api.route("/get_classes", methods=["GET"])
+@with_session
+def get_classes_by_teacher_id():
```
mircealungu: ditto about `class` vs `cohort`.

there was a lot of thinking actually in the 
naming of the cohort class. although it 
sounds a bit less usual, it can't be confused
with the programming language concept, 
and that can be handy in more obscure parts
of the code.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+        for link in links:
+            db.session.delete(link)
+        db.session.delete(selected_cohort)
+        db.session.commit()
+        return 'removed'
+    except ValueError:
+        flask.abort(400)
+
+
+# Takes Teacher id as input and outputs list of all cohort_ids that teacher owns
```
mircealungu: if you say "as input" i think that either the method
or the endpoint has the id as a parameter. but this
is not the case. 

it rather **returns the cohorts for which the currently logged in user is a teacher**. Also, this would imply that `get_own_cohorts` would be a nicer name.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+    mappings = TeacherCohortMap.query.filter_by(user_id=flask.g.user.id).all()
+    cohorts = []
+    for m in mappings:
+        info = _get_class_info(m.cohort_id)
+        cohorts.append(info)
+    return json.dumps(cohorts)
+
+
+# Takes cohort_id and reuturns dictionary with relevant class variables
+@api.route("/get_class_info/<id>", methods=["GET"])
```
mircealungu: same problem as before ... it does not take the teacher id... it seems to take the cohort id. 
also, the GET is in the HTTP method name so you can drop that.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+@with_session
+def wrapper_to_json_class(id):
+    if (not _has_permission_for_class(id)):
+        flask.abort(401)
+    return jsonify(_get_class_info(id))
+
+
+def _get_class_info(id):
```
mircealungu: cohort

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+@with_session
+def check_inv_code(invite_code):
+    c = Cohort.query.filter_by(inv_code=invite_code).first()
+    if c is None:
+        return jsonify(1)
+    return jsonify(0)
+
+
+# creates a class in the data base. Requires form input (inv_code, class_name, class_language_id, max_students, teacher_id)
+@api.route("/add_class", methods=["POST"])
```
mircealungu: `/add_own_cohort` would make it even more explicit that 
the cohort is added to the logged in user. 

It might be possible that in the future, another user, a super user
or admin would be able to `/create_cohort` for a teacher. thus, 
the `own` would make a difference.

zeeguu_api/api/dashboard.py
mircealungu: also, before starting to work on Human Readable error messages ask @LarsHoldijk  to share with you his Modification Proposal that talks about errors and also maybe have a discussion with him!

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+
+# Get user bookmarks
+@api.route("/get_user_bookmarks/<id>", methods=["GET", "POST"])
+@with_session
+def get_user_bookmarks(id):
+    user = User.query.filter_by(id=id).one()
+    if (not _has_permission_for_class(user.cohort_id)):
+        flask.abort(401)
+    # True input causes function to return context too.
+    return json_result(user.bookmarks_by_day(True))
```
mircealungu: i alway forget what's the difference between this one
and bookmarks by day. could we name it `cohort_member_bookmarks`? 
and add a comment that mentions that this is different than the other
one because it's used from the perspective of the teacher.

finally, why is POST a possible method here? it clearly should only be GET.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+            flask.abort(400)
+
+        class_to_change.max_students = request.form.get("max_students")
+
+        db.session.commit()
+        return 'updated'
+    except ValueError:
+        flask.abort(400)
+        return 'failed'
+    except sqlalchemy.exc.IntegrityError:
```
mircealungu: well handled exceptions.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+        class_to_change.inv_code = request.form.get("inv_code")
+        class_to_change.class_name = request.form.get("class_name")
+
+        if int(request.form.get("max_students")) < 1:
+            flask.abort(400)
+
+        class_to_change.max_students = request.form.get("max_students")
+
+        db.session.commit()
+        return 'updated'
```
mircealungu: if we'll agree with the "OK" for success, then we should return that here

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+@with_session
+def remove_class(class_id):
+    from zeeguu.model import TeacherCohortMap
+    if (not _has_permission_for_class(class_id)):
+        flask.abort(401)
+    try:
+        selected_cohort = Cohort.query.filter_by(id=class_id).one()
+
+        if not selected_cohort.cur_students == 0:
+            flask.abort(400)
```
LarsHoldijk: The client should be notified that the cohort couldn't be removed because of it still having students.

zeeguu_api/api/dashboard.py
Oli26: Do you want me to also rename the variables in cohort from class_name to cohort_name etc?
This might take me a while to do as it requires changes from many functions in the API and the WEB

mircealungu: Sure. Nice idea. Or name it only "name" given that it.s already scoped in cohort?

Oli26: All API funcitons have been renamed accordingly. All variable names in Cohort have been renamed accordingly. All tests have been updated to accommodate for changes. I might save the function names as variables in testing in future to make test updating faster.

zeeguu_api/api/dashboard.py
Oli26: Removed all the "get"s from function calls! (unless they are not public to the rest api)

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+# Checks to see if user has a valid active session
+@api.route("/has_session", methods=["GET"])
+def has_session():
+    try:
+        session_id = int(flask.request.args['session'])
+        session = Session.query.filter_by(id=session_id).one()
+        return jsonify(1)
+    except:
```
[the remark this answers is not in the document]
Oli26: I have tried to update dashboard.py so that call its functions catch the correct exceptions.

tests_zeeguu_api/test_dashboard.py
```diff
@@ -0,0 +1,277 @@
...
+from zeeguu_api.app import app
+
+
+class DashboardTest(APITestMixin, TestCase):
+
+    def test_has_session(self):
+        # TEST FOR FAKE SESSION
+        result = self.app.get('/has_session?session=31235ds')
+        result = json.loads(result.data)
+        assert result == 0
```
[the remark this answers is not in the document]
Oli26: Changed my success returns to "OK"!

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,365 @@
...
+db = zeeguu.db
+
+
+def _has_permission_for_cohort(cohort_id):
+    '''
+        Checks to see if user requesting has permissions to view the cohort with id 'cohort_id'
+    '''
+    from zeeguu.model import TeacherCohortMap
+    links = TeacherCohortMap.query.filter_by(cohort_id=cohort_id).all()
+    for l in links:
```
mircealungu: still don't like this for l in links.
i think later you use `for m in maps` which is slightly better.

mircealungu:
once the last changes are done, we should be able to merge your changes in the master! 
thanks for the effort @Oli26  and @evixhelo

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,365 @@
...
+@api.route("/check_invite_code/<invite_code>", methods=["GET"])
+@with_session
+def check_inv_code(invite_code):
+    '''
+        Checks if the inputted invite code is already in use.
+
+    '''
+    c = Cohort.query.filter_by(inv_code=invite_code).first()
+    if c is None:
+        return jsonify(1)
```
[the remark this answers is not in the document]
Oli26: Sorry I thought I changed this but must have missed it.

zeeguu_api/api/dashboard.py
```diff
@@ -0,0 +1,365 @@
...
+    links = TeacherCohortMap.query.filter_by(cohort_id=cohort_id).all()
+    for l in links:
+        if l.user_id == flask.g.user.id:
+            return True
+    return False
+
+
+@api.route("/has_session", methods=["GET"])
+def has_session():
+    '''
```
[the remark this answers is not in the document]
Oli26: I didn't fully comprehend your explaination before but I just checked the validate function with postman and I get it now. Will remove this and change the WEB to use Validate()!

PR #23, opened 2018-04-23
Dashboard This should be the correct one!

Oli26:


zeeguu_api/api/teacher_dashboard.py
```diff
@@ -224,8 +223,8 @@ def check_inv_code(invite_code):
     '''
     c = Cohort.query.filter_by(inv_code=invite_code).first()
     if c is None:
-        return jsonify(1)
-    return jsonify(0)
+        return "Taken"
```
mircealungu: A slightly more informative error wouldn't hurt ;)

mircealungu:
Fair enough.

M.

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -215,17 +214,17 @@ def _link_teacher_cohort(user_id, cohort_id):
...
+def inv_code_exists(invite_code):
     '''
         Checks if the inputted invite code is already in use.
 
     '''
     c = Cohort.query.filter_by(inv_code=invite_code).first()
     if c is None:
-        return jsonify(1)
-    return jsonify(0)
+        return "False"
```
mircealungu: False -- is not the best answer, i'm sorry :) 
think of the implications: 

```
Q: /invite_code_exists?
A: False

```
I guess /validate_invite_code 

would be ok to have False as a result.

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -31,7 +31,7 @@ def has_permission_for_cohort(cohort_id):
 
 @api.route("/has_permission_for_cohort/<id>", methods=["GET"])
 @with_session
-def has_permission_for_cohorts(id):
+def has_permission_for_cohort_public(id):
```
mircealungu: what does public mean here?

Oli26: There is already a function has_permissions_for_cohort that is private and used inside other functions. Whereas the routed has_permissions_for_cohort is mainly used on the dashboard to identify if it should attempt to load something from the API.

mircealungu:
Oh!
But then can't you use the underscore in front of the private method names
python convention?

E.g. _fibonacci vs fibonacci

M.

PR #25, opened 2018-04-28
Dashboard

Oli26:


zeeguu_api/api/teacher_dashboard.py
```diff
@@ -105,14 +106,33 @@ def _get_user_info(id):
...
             'name': user.name,
             'email': user.email,
-            'reading_time': random.randint(1, 100),
-            'exercises_done': random.randint(1, 100),
-            'last_article': 'place holder article'
+            'reading_time': reading_time,
+            'exercises_done': exercise_time,
+            'last_article': 'place holder article',
+            'reading_time_list': reading_time_list,
+            'exercise_time_list': exercise_time_list
```
mircealungu: this part of the code must be better documented and discussed with @carlospaz7 . 
eventually he is the one that computes the information about Reading Sessions and Exercise Sessions.
what is the reading_time_list? is it a list of sessions? 
and the `reading_time` is the total reading time?

mircealungu: also, shouldn't this method also work only on a given interval?

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -304,17 +324,26 @@ def add_user_with_cohort():
     flask.abort(400)
 
 
-@api.route("/cohort_member_bookmarks/<id>", methods=["GET"])
+@api.route("/cohort_member_bookmarks/<id>/<time_period>", methods=["GET"])
 @with_session
-def cohort_member_bookmarks(id):
+def cohort_member_bookmarks(id, time_period):
     '''
         Returns books marks from member with input user id.
```
mircealungu: bookmarks

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -94,7 +95,7 @@ def wrapper_to_json_user(id):
         then returns result jsonified.
```
mircealungu: takes the id of a user, right?

zeeguu_api/api/teacher_dashboard.py
Oli26: This is integrated now and should made more sense. It returns a list for a table and the total time for the bars for each student. Now includes a duration for which data to return.

PR #30, opened 2018-06-21
Caching

Feikoritsema:
**Warning: this requires the new core and database** 

When removing a source, no methods are called in the core, so we must update the hash for the caching from the API.

### DIFF
--- a/zeeguu_api/api/feeds.py
+++ b/zeeguu_api/api/feeds.py
@@ -2,7 +2,7 @@ import flask
 import zeeguu
 from flask import request
 from zeeguu.model import RSSFeedRegistration, RSSFeed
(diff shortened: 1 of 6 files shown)


zeeguu_api/api/feeds.py
```diff
@@ -2,7 +2,7 @@
 import zeeguu
 from flask import request
 from zeeguu.model import RSSFeedRegistration, RSSFeed
-from zeeguu.content_recommender.mixed_recommender import recompute_recommender_cash_if_needed
```
Feikoritsema: Wow I see what you ment now, I'm embarrassed

mircealungu:
make sure to pull this too: i've renamed the method in the API to match the new name in the Core.

PR #42, opened 2019-03-11
Teacher dashboard

jsparvath:
Added methods enabling seeing the users reading_session activity. This allows us to see the user activity even if there are no bookmarks related to the session

### DIFF
--- a/zeeguu_api/api/teacher_dashboard.py
+++ b/zeeguu_api/api/teacher_dashboard.py
@@ -168,26 +168,6 @@ def _get_user_info_for_teacher_dashboard(id, duration):
     except ValueError:
         flask.abort(400)
         return 'ValueError'
-
-@api.route("/user_reading_sessions/<id>/<duration>", methods=["GET"])
-@with_session
-def get_user_reading_sessions(id, duration):
-    '''
-        Returns a serialized list of dictionaries of the sessions belonging to the user.
-
-    '''
-    from zeeguu_core.model import UserReadingSession
-    user = User.query.filter_by(id=id).one()
-    if (not has_permission_for_cohort(user.cohort_id)):
-        flask.abort(401)
-    try:
-        fromDate = datetime.now() - timedelta(days=int(duration))
-        reading_sessions = UserReadingSession.find_by_user(int(id), fromDate, datetime.now())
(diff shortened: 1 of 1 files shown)


zeeguu_api/api/teacher_dashboard.py
```diff
@@ -272,6 +271,20 @@ def _link_teacher_cohort(user_id, cohort_id):
...
+    '''
+        Return list of dictionaries containing user info for all cohorts that the logged in user owns.
+
+    '''
+    from zeeguu_core.model import TeacherCohortMap
+    mappings = TeacherCohortMap.query.filter_by(user_id=flask.g.user.id).all()
+    all_users = []
+    for m in mappings:
+        users = users_from_cohort(m.cohort_id, duration)
+        all_users.extend(json.loads(users))
```
mircealungu: i'm a bit puzzled by this. 

are you sure that converting to json and then deconverting 
again here is the nicest way to solve the problem you have?
i wonder if you could refactor the other function to extract a local 
that you can reuse both here and there ... something like: 

```
user_info_from_cohort
        c = Cohort.query.filter_by(id=id).one()
        users = User.query.filter_by(cohort_id=c.id).all()
        users_info = []
        for u in users:
            info = _get_user_info_for_teacher_dashboard(u.id, duration)
            users_info.append(info)
       return users_info
```

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -343,6 +356,28 @@ def add_teacher(id):
...
+        flask.abort(400)
+        return "NoUserFound"
+
+    if (not has_permission_for_cohort(user.cohort_id)):
+        flask.abort(401)
+
+    now = datetime.today()
+    date = now - timedelta(days=int(time_period));
+
+    # True input causes function to return context too.
```
mircealungu: i can't parse this comment. what does it mean?

jsparvath: Ah, it was a leftover from the method I borrowed. removed it

zeeguu_api/api/teacher_dashboard.py
jsparvath: Good point. much cleaner now :)

PR #49, opened 2019-04-01
Upload / deletion of articles for teacher dashboard

mbrodt:
This pull request adds 2 new endpoints to the API:
1. An "upload file" endpoint which takes some information about an article, and creates it in the database. This function also creates a map between the cohort where the article was added, and the article id (this is to only display articles for a given class on a class page). 

2. An endpoint to delete previously uploaded articles if the person attempting deletion has access to the cohort associated with the article being deleted.

### DIFF
--- a/zeeguu_api/api/teacher_dashboard.py
+++ b/zeeguu_api/api/teacher_dashboard.py
@@ -438,23 +438,23 @@ def update_cohort(cohort_id):
         return 'IntegrityError'
 
 
-@api.route("/upload_files/<cohort_id>", methods=["POST"])
+@api.route("/upload_articles/<cohort_id>", methods=["POST"])
 @with_session
-def upload_file(cohort_id):
+def upload_articles(cohort_id):
     '''
-        uploads a file for a cohort with input from a POST request
+        uploads articles for a cohort with input from a POST request
     '''
     if (not has_permission_for_cohort(cohort_id)):
         flask.abort(401)
     try:
-        for file in json.loads(request.data):
-            url = Url('teacher-dash-file/{}'.format(datetime.now()))
-            title = file['title']
-            authors = file['authors']
-            content = file['content']
-            summary = file['summary']
+        for article_data in json.loads(request.data):
+            url = Url('userarticle/{}'.format(uuid.uuid4().hex))
+            title = article_data['title']
+            authors = article_data['authors']
+            content = article_data['content']
+            summary = article_data['summary']
             published_time = datetime.now()
-            language_code = file['language_code']
+            language_code = article_data['language_code']
             language = Language.find(language_code)
 
             new_article = Article(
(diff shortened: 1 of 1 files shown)


zeeguu_api/api/teacher_dashboard.py
```diff
@@ -432,3 +436,81 @@ def update_cohort(cohort_id):
...
+@api.route("/upload_files/<cohort_id>", methods=["POST"])
+@with_session
+def upload_file(cohort_id):
+    '''
+        uploads a file for a cohort with input from a POST request
+    '''
+    if (not has_permission_for_cohort(cohort_id)):
+        flask.abort(401)
+    try:
+        for file in json.loads(request.data):
```
mircealungu: file used to be a builtin... also, technically you're not getting a file but rather an `article_data` it seems to me

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -432,3 +436,81 @@ def update_cohort(cohort_id):
...
+            url = Url('teacher-dash-file/{}'.format(datetime.now()))
```
mircealungu: I propose `Url('user-article-{}'.format(...)`. what do you think?

mircealungu: also, that will have spaces in the name... 
e.g. 

    'user-article-2019-04-02 14:38:07.416790'

is this what you indended?

mircealungu: maybe something like this?

    'userarticle/{}'.format(uuid.uuid4().hex)

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -432,3 +436,81 @@ def update_cohort(cohort_id):
...
+                content,
+                summary,
+                published_time,
+                None,  # rss feed
+                language
+            )
+            
+            db.session.add(new_article)
+            db.session.flush()
+            db.session.refresh(new_article)
```
mircealungu: this is interesting. 
why do add+flush+refresh?

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -432,3 +436,81 @@ def update_cohort(cohort_id):
...
+
+
+@api.route("/cohort_files/<cohort_id>", methods=["GET"])
+@with_session
+def cohort_files(cohort_id):
+    '''
+        Gets the files associated with a cohort
+    '''
+    cohort = Cohort.find(cohort_id) 
+    articles = CohortArticleMap.get_articles_for_cohort(cohort)
```
mircealungu: make sure to refactor this if you rename the function in core!

zeeguu_api/api/teacher_dashboard.py
```diff
@@ -432,3 +436,81 @@ def update_cohort(cohort_id):
...
+    return json.dumps(articles)
+
+@api.route("/delete_article/<cohort_id>/<article_id>", methods=["POST"])
+@with_session
+def delete_article(cohort_id, article_id):
+		'''
```
mircealungu: in the GH diff UI there's a lot of indentation in this method.
can you check why? 

Screen Shot 2019-04-02 at 2 47 20 PM

mbrodt: I tried to fix this, but in the code it looks normal. Let me know if its still happening

mircealungu:
I see tests are passing, and you clarified my last question. I'm merging this into master.

PR #51, opened 2019-04-09
Add lazy fetching of translations

alinbalutoiu:


alinbalutoiu:
@mircealungu the API now exposes ``get_next_translations`` which works both for get_top_translations and get_rest_translations. You just need to pass the services that you want to exclude as parameters. Can you have a look over the PR and tell me your thoughts?

PR #53, opened 2019-04-11
added cohort article endpoint

jsparvath:


zeeguu_api/api/user_articles.py
```diff
@@ -33,6 +33,22 @@ def user_articles_starred_and_liked():
...
+def user_articles_cohort(count: int = 20):
+    """
+        get articles for the cohort associated with the user
+    """
+
+    return json_result(cohort_articles_for_user(flask.g.user, count))
+
+
+
+
```
alinbalutoiu: Too many blank lines, we should try to stick with the pep8 style guidelines, usually you have 2 blank lines between different methods.

mbrodt:
Hey guys, we've fixed the comments and changed the API so it now removes the article from a cohort, rather than actually deleting an article (like we discussed with @mircealungu)

Would appreciate a merge :)

mircealungu:
I've merged the core.
Can you @mbrodt make another minor change such that the tests pass? Then I can merge.

mbrodt:
Hm, I'm not quite sure what the problem is. The tests give this error:

"ImportError: cannot import name 'cohort_articles_for_user'"

which doesnt make sense to me, as the code works fine and has no problem importing the function from core.

mircealungu:
is this the context in which i said that
you should make a dummy change ?
it might be that the tests were run with
the old core. now i've merged your changes
in the core, and if they would be run again
they would be ok. however, for them to be
automatically run you need to push a new
commit...

mbrodt:
Aha, I see - you were indeed correct 👍 
 
Would it be possible to deploy this as well as the core @mircealungu ?

PR #95, opened 2023-05-17
Updates to nlp components

tfnribeiro:
# These changes are made to help support the logging activity.

## nlp
- create_confusion_words: Instead of just providing the confusion words the dictionary is returned, and the frontend can use all the properties. This was done to support the user_activity
- added a new endpoint get_sentences_for_wo, which should implement the logic to find sentences when WO has a very long sentence. Currently, the highest similarity sentence with a limit of 15 tokens is used.

## automatic_gec_tagging
- Fixed the article, it was checking the wrong parameter to select the article used.

## confusion_generator
- Changed the names in the dictionary to make more sense.

## spacy_wrapper
- Added a method to get similarity of two sentences.

### DIFF
--- a/zeeguu/api/api/nlp.py
+++ b/zeeguu/api/api/nlp.py
@@ -7,6 +7,7 @@ from flask import request
 from zeeguu.core.nlp_pipeline import SpacyWrappers, NoiseWordsGenerator
 from zeeguu.core.nlp_pipeline import AutoGECTagging
 
+import heapq
 # ---------------------------------------------------------------------------
 @api.route("/do_some_spacy", methods=("POST",))
 # ---------------------------------------------------------------------------
@@ -38,7 +39,7 @@ def create_confusion_words():
         return "Language not supported"
     
     # We should pass the student bookmark words as a fallback when no words are found.
-    noise_words = NoiseWordsGenerator[language].generate_confusion_words(original_sent)["confusion_words"]
+    noise_words = NoiseWordsGenerator[language].generate_confusion_words(original_sent)
     return json_result(noise_words)
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


zeeguu/api/api/nlp.py
```diff
@@ -58,3 +59,28 @@ def annotate_clues():
 
     return json_result(updated_words)
 
+# ---------------------------------------------------------------------------
+@api.route("/get_sentences_for_wo", methods=("POST",))
+# ---------------------------------------------------------------------------
+@cross_domain
+@with_session
+def get_sentences_for_wo():
```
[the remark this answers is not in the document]
tfnribeiro: Currently, the way I implemented it is within the OrderWords exercise itself - so if the context is too long, then the articleInfo is used to find a new sentence that is smaller. So at that point, we only get the article associated with that bookmark. However, the issue with this is the fact that they might not see the word again, so we would have to change the outcome to not update the tries with the word associated. 

I am hoping the method that just selects a subsentence works well because then it would avoid this type of cases. 

In sum:
Inputs: Article Text, Bookmark Context, Language
Outputs: Heap with the Similarities based on the spaCy embeddings [(similarity, Sentence), ...]

zeeguu/api/api/nlp.py
```diff
@@ -58,3 +59,28 @@ def annotate_clues():
...
+    article_text = request.form.get("article_text", "")
+    bookmark_context = request.form.get("bookmark_context", "")
+    language = request.form.get("language")
+
+    if language not in SpacyWrappers.keys():
+        return "Language not supported"
+    
+    nlp_pipe = SpacyWrappers[language]
+    sentences = nlp_pipe.get_sent_list(article_text)
+    filtered_sentences = [sent for sent in sentences if len(nlp_pipe.tokenize_sentence(sent)) <= 15]
```
mircealungu, line 77: rename to `short_sentences_in_article`?

zeeguu/api/api/nlp.py
```diff
@@ -58,3 +59,28 @@ def annotate_clues():
...
+    
+    context_doc = nlp_pipe.spacy_pipe(bookmark_context)
+    heap = []
+    for f_sent in filtered_sentences:
```
mircealungu, line 81: I see what's confusing to me now.

`nlp_pipe.spacy_pipe` sounds confusing. Could we maybe add a `process()` function in the SpacyWrapper, and optionally name the var above `spacy_pipe` instead of `nlp_pipe` such that the code will look like: `context_doc = spacy_pipe.process(bookmark_context)`? 

That way this code will read easier I think.

zeeguu/api/api/nlp.py
```diff
@@ -58,3 +59,28 @@ def annotate_clues():
...
+        sent_doc = nlp_pipe.spacy_pipe(f_sent)
+        heapq.heappush(heap, (-context_doc.similarity(sent_doc), f_sent))
+
+    _, most_similar_sent = heap.pop()
```
mircealungu, line 85: so you're throwing away the most similar sentence - i guess that's the bookmark sentence? to make it clear maybe simply do `_,_ = heap.pop()`

mircealungu (approved):

zeeguu/api/api/nlp.py
tfnribeiro: There was actually a bug here. I was trying to put this change in order to allow us to check results directly in the console without having to push changes again. 

So I wanted to return the top sentence, but also the rest of the sentences in the article, I made it more clear now, and instead, I send a json with the top 1 and top 10 sentences (with the similarity).

mircealungu (approved):

PR #100, opened 2023-12-01
updated plain text pattern for politiken

mircealungu:
tiago, did I understand your email correctly? 
if we filter articles that contain this word we have a good approximation for Danish paywalled articles?


mircealungu:
oki. 
let’s see!
i wonder if you can beat my heuristic ;)

m.

PR #108, opened 2024-01-27
supporting a live reader activity timer

mircealungu:
Updated the backend to support a live timer for the reader


tfnribeiro (approved):
I see we will now have a `endpoints/helpers/` folder. However, we already had a `api/utils `folder. I am assuming we will only place rather specific endpoint related utility folders in this new folder, but I could also see a point in maybe keeping it under `utils`? What do you think?

mircealungu:
> I see we will now have a `endpoints/helpers/` folder. However, we already had a `api/utils `folder. I am assuming we will only place rather specific endpoint related utility folders in this new folder, but I could also see a point in maybe keeping it under `utils`? What do you think?

Good point! 

At the first glance the `api/utils` are more meta-level: they define wrappers, etc. What I added in `/helpers` is more common code. Let's keep it for now, and keep an eye open for hints that it should be merged or not as we enrich the api in the upcoming period?

PR #111, opened 2024-02-01
78 text sources

tfnribeiro:
Added support for the new feed types. Currently, we can use the RSS feed (the previous solution) and use the newspaper library (0 and 1, respectively). 

I am attaching the checklist that I used to make sure the changes and sources worked. 

#### 1. Drop and Recreate zeeguu_test schema

✅ DONE, no issues.

#### 2. Load the DB dump sent from the 29th of January 2024

✅ DONE, no issues.

#### 3. Run pytest from the API master code (docker-container)

✅ DONE, no issues. Test Results: 161 passed, 28888 warnings in 109.61s (0:01:49)

#### 1. Run the migration/rss_to_feed.sql in DB

✅ DONE, updated the script to seperate the ALTER TABLE statements for rss_feed and running it on mysql shows no errors.

#### 2. Run pytest from the API master code (docker-container)

✅ DONE, pytest runs successfuly. 

161 passed, 28165 warnings in 121.80s (0:02:01). I would exepect errors (?), but none were seen. Is it because of the mock code? We use a in-memory model, which means that the code doesn't run into errors.

Logging into the app results in errors: ```sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1054, "Unknown column 'article.rss_feed_id' in 'field list'")``` (as expected)

#### 3. Switch to update branch (78-text-sources)

✅ DONE, no issues.

#### 4. Run pytests using the updated branch

✅ DONE, no issues. Test Results: 163 passed, 30690 warnings in 173.78s (0:02:53)
+2 tests were added related to the newspaper change. 

#### 5. Log into application and check for errors

⚠ DONE, no errors but my test user (Set for Danish) is retrieving Dutch and French articles. Could be the Elastic search?

✅ Crawling the articles resulted in the ElasticSearch showing the Danish articles. 

#### 6. Crawl for new sources and see if the previous error disapears.

✅ DONE, crawling articles removes issues seen in 5. (Elastic search needs to be re-indexed to the new DB dump).

#### 7. Add sources (newspaper) and crawl them.

   7.1. https://www.seoghoer.dk/ 

   ✔ DONE, Found 36 articles.

   7.2. https://verdensbedstenyheder.dk/

   ✔ DONE, Found 38 articles.

   7.3. https://uniavisen.dk/

   ✔ DONE, Found 19 articles.

   7.4. https://magasinet360.dk/

   ✔ DONE, Found 23 articles.

All of these links emit a warning when trying to access the RSS feed.

#### 8. Download the articles using 'article_crawler.py'  (with a timeout of 120 seconds) - Crawled 2 days

   8.1. Downloaded 59 articles (no low quality)

   8.2. Downloaded 37 articles (no low quality)

   8.3. Downloaded 17 articles (no low quality)

   8.4. Downloaded 23 articles (no low quality)

Some timeouts from the ElasticSearch, probably due to local dev environment.
Looking at the crawler download in Danish, I can notice that a lot of articles are not assigned topics

#### 9. Merge latest API changes and see if the code still works.

✅ DONE, one merge conflict (UserReadingSession) the file was removed.

#### 10. Re-run tests:

✅ DONE, 147 passed, 27537 warnings in 140.07s (0:02:20)

#### 11. Use the application:

✅ DONE, could find articles from the new sources. The crawler seems to have extracted the text well. No other errors seen. 

# Old:

1. Run the migration/rss_to_feed.sql in DB

⚠ Despite the initial error, it seems to work if each statement is run individually and in order. 

❌ ERROR! When running the DB I run into this error:

```ALTER TABLE `zeeguu_test`.`rss_feed` 
 ADD COLUMN `feed_type` INT NOT NULL DEFAULT 0 AFTER `deactivated`, RENAME TO  `zeeguu_test`.`feed`	```
 
 Error Code: 1050. Table 'feed' already exists	0.000 sec
	
```ALTER TABLE `zeeguu_test`.`article` 
ADD CONSTRAINT `article_ibfk_1`
   FOREIGN KEY (`feed_id`)
   REFERENCES `zeeguu_test`.`feed` (`id`)```

Error Code: 1452. Cannot add or update a child row: a foreign key constraint fails (`zeeguu_test`.`#sql-1cf0_1d`, CONSTRAINT `article_ibfk_1` FOREIGN KEY (`feed_id`) REFERENCES `feed` (`id`))	0.016 sec

This seems to happen because when the RENAME fails, then the constraint cannot be created.

⚠ DONE (with WARNING), ran each part of the script individually `ALTER TABLE statements for 'rss_feed'` and then the `ALTER TABLE for 'article'`.

```ALTER TABLE `zeeguu_test`.`article` 
 ADD CONSTRAINT `article_ibfk_1`
   FOREIGN KEY (`feed_id`)
   REFERENCES `zeeguu_test`.`feed` (`id`)```
   
   Error Code: 2013. Lost connection to MySQL server during query	30.016 sec

✅ I ran this from Workbech, which has a 30 second timeout for queries. When I log back into the database, I can see that the article table has all the values updated. Looks OK otherwise.

### CODE REVIEW

File: tools/migrations/add_icon_name_to_rss_feed.py [hunk 1]
```diff
@@ -0,0 +1,12 @@
+import zeeguu.core
```
mircealungu, line 1: why is this needed?
it is not referenced in the checklist?

tools/migrations/rss_to_feed.sql
```diff
@@ -0,0 +1,15 @@
...
+*/
+ALTER TABLE `zeeguu_test`.`rss_feed` RENAME TO  `zeeguu_test`.`feed`;
+ALTER TABLE `zeeguu_test`.`feed` ADD COLUMN `feed_type` INT NOT NULL DEFAULT 0 AFTER `deactivated`;
+
+/* Update the article table to match the updated feed table.
+*/
+ALTER TABLE `zeeguu_test`.`article` DROP FOREIGN KEY `article_ibfk_1`;
+ALTER TABLE `zeeguu_test`.`article` CHANGE COLUMN `rss_feed_id` `feed_id` INT NULL DEFAULT NULL ;
+ALTER TABLE `zeeguu_test`.`article` 
+ADD CONSTRAINT `article_ibfk_1`
```
mircealungu, line 13: i wonder if there is a better naming of fk constraints. 
this conversation seems to suggest that article_fk_feed_id? https://dba.stackexchange.com/questions/15530/what-does-ibfk-stand-for-in-mysql
then again, this would only apply for the FKs that are defined by us. the ones generated seem to indeed of of the form ibfk_<number>

zeeguu/core/content_retriever/parse_with_readability_server.py
```diff
@@ -6,6 +6,8 @@
 
 
 def download_and_parse(url):
+    # This code will be run twice when using a newspaper source
+    # Maybe add a flag to avoid running this if the feed type == newspaper
```
mircealungu, line 10: why is it running twice? how can we avoid it?

zeeguu/core/feed_handler/feed_handler.py
```diff
@@ -0,0 +1,27 @@
...
+
+class FeedHandler():
+    def __init__(self, url:str, feed_type:int):
+        self.url = url
+        self.feed_type = feed_type
+        self.title = ""
+        self.description = ""
+        self.image_url_string = ""
+
+    def get_utc_time(self, article_date) -> datetime:
```
mircealungu, line 12: ah. timezones again!
why is it called `get_utc_time` if it gets current time on the server? 
should it not be called `get_server_time`?

zeeguu/core/feed_handler/newspaperfeed.py
```diff
@@ -0,0 +1,54 @@
...
+            published_datetime:datetime, date time of the article
+        """
+        print("Newspaper Built!")
+        # Not sure if we should use cache (as currently the crawler checks if the article is in)
+        # This makes it complicated to assign a feed and download the articles found at that time.
+        # Currently, it ignores the newspaper's cache and justs uses ours.
+        if self.use_cache:
+            news_feed = newspaper.build(self.url)
+        else:
+            news_feed = newspaper.build(self.url, memoize_articles=False)
```
mircealungu, line 35: nice and clean.
i like it!

zeeguu/core/feed_handler/newspaperfeed.py
```diff
@@ -0,0 +1,54 @@
...
+from zeeguu.logging import log
+
+
+class NewspaperFeed(FeedHandler):
+    def __init__(self, url: str, feed_type: int, use_cache: bool = True):
+        super()
+        print("Created Newspaper Source")
+        self.url = url
+        self.feed_type = feed_type
+        data = newspaper.Source(url)
```
mircealungu, line 13: I wonder if it would make sense to move this in a separate function named `extract_feed_metadata`. 
there's a parallel also in the RSSFeed handler. 
And the two strategies are different. 
then in the superclass, you should be able to initialize the url, feed_type, use_cache, and then simply call `self.extract_feed_metadata` that will run the corresponding implementatin ?

zeeguu/core/util/time.py
```diff
@@ -0,0 +1,21 @@
...
+    # to UTC.
+    return (
+        datetime.now()
+        .astimezone(timezone(timedelta(hours=SERVER_HOUR_DIFFERENCE)))
+        .astimezone(timezone.utc)
+        .replace(tzinfo=None)
+    )
+
+
+def normalize_to_server_time(date: datetime):
```
mircealungu, line 18: comment pls.

zeeguu/core/util/time.py
```diff
@@ -0,0 +1,21 @@
+from datetime import datetime, timezone, timedelta
+
+
+SERVER_HOUR_DIFFERENCE = 1  # Frankfurt is +1 from UTC
+
+
+def get_server_time_utc():
+    # Returns the time with the time delta of the server, converted
```
mircealungu, line 8: i officially hate timezones.
can you detail a bit more in the comment why do we need this?

mircealungu (approved):
a few comments left in there for you @tfnribeiro

tools/migrations/add_icon_name_to_rss_feed.py
tfnribeiro: I think this might have been one of the migrations we deleted, but I still had it tracked in my branch. I think we can remove it if you also agree (or move it to the old folder?)

tools/migrations/rss_to_feed.sql
tfnribeiro: I think that would mean updating all the FKs to a new naming scheme, as currently we always use [table]_ibfk_ . So, I would say that should be done as a different change?

zeeguu/core/feed_handler/feed_handler.py
tfnribeiro: Correct, it should be called get_server_time , I will update it.

zeeguu/core/content_retriever/parse_with_readability_server.py
tfnribeiro: When we crawl using newspaper we already download the articles, so it we already have the parsed information. I thought about adding a parameter to the call of the method: `feed_type`, then we can handle the fact that the newspaper text is already available.

tools/migrations/add_icon_name_to_rss_feed.py
mircealungu: ah. that makes sense. let's delete it then!

zeeguu/core/util/time.py
tfnribeiro: I have added a long doc string explaining what we discussed. It's a bit hard to put it in a small amount of words. I recently read an article in datetime and why it's a bit of a pain, and I can relate writing this.

The main point is the tz awareness of datetime objects. 

`get_server_time_utc` is unused at the moment, it's in there in case we ever need to convert the server time to UTC and have it be tz aware for a certain operation with other tz objects.

zeeguu/core/content_retriever/parse_with_readability_server.py
tfnribeiro: However, we have multiple places in the code where this is called without having the feed information naturally. So I am not sure if adding a parameter would make sense. This method is essentially asking for an URL and returning a newspaper.Article object. So it might be difficult to say that we already have that object at all times.

zeeguu/core/feed_handler/newspaperfeed.py
tfnribeiro: I have added the method extract_feed_metadata to the interface and implemented them each in the sub-classes. So now, when an object is created they call their super constructor and then extract the metadata as you suggested 👍

zeeguu/core/feed_handler/newspaperfeed.py
```diff
@@ -0,0 +1,55 @@
...
+from .feed_handler import FeedHandler
+
+from zeeguu.logging import log
+
+
+class NewspaperFeed(FeedHandler):
+    def __init__(self, url: str, feed_type: int, use_cache: bool = True):
+        super().__init__(url, feed_type)
+        self.use_cache = use_cache
+        self.extract_feed_metadata()
```
mircealungu, line 11: Given that the `extract_feed_metadata` is called in both the subclass inits, it would make sense to move them to the superclass. Something like: 

```
self.url = url
         self.feed_type = feed_type
         self.title = ""
         self.description = ""
         self.image_url_string = ""
         self. extract_feed_metadata()
```

and then this one becomes

```
    def __init__(self, url: str, feed_type: int, use_cache: bool = True):
         self.use_cache = use_cache
         super().__init__(url, feed_type)
```

altough, it's a very very minor observation!

mircealungu (approved):

zeeguu/core/feed_handler/newspaperfeed.py
tfnribeiro: Oh that is a good point, since it wasn't implemented in the Super-class I ignored it, but let's do it that way. It's much cleaner!

mircealungu:
@tfnribeiro - I guess this one got a bit stale :) 
is it ready to merge?

tfnribeiro:
@mircealungu It should be ready! I made all the changes and it was looking good. Maybe doing a test deployment in your dev environment would be good just to be sure.

PR #114, opened 2024-02-13
Adding end-point to remove saved articles

tfnribeiro:
Added an end-point similar to `make_personal_copy` called `remove_personal_copy` to delete saved articles. 

This end-point can be called from the Web to remove the saved articles.

### DIFF
--- a/zeeguu/api/endpoints/article.py
+++ b/zeeguu/api/endpoints/article.py
@@ -67,6 +67,7 @@ def make_personal_copy():
 
     return "OK" if PersonalCopy.exists_for(user, article) else "Something went wrong!"
 
+
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


zeeguu/api/endpoints/article.py
```diff
@@ -68,6 +68,25 @@ def make_personal_copy():
...
+
+    article_id = request.form.get("article_id", "")
+    article = Article.find_by_id(article_id)
+    user = flask.g.user
+
+    if PersonalCopy.exists_for(user, article):
+        PersonalCopy.remove_for(user, article, db_session)
+
+    return (
+        "OK" if not PersonalCopy.exists_for(user, article) else "Something went wrong!"
```
mircealungu, line 86: nice and pythonic :)

PR #116, opened 2024-02-14
Match Workflow python version with deployed version

tfnribeiro:
From the docker deployment file:

`FROM python:3.9.2-buster`

Shouldn't we ensure the version is the same on the server? We could also upgrade python in the docker to the latest version? 

https://devguide.python.org/versions/ Seems like 3.9 ends in 2025, and since no errors are seen for 3.11 maybe we could just upgrade it?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


mircealungu (approved):
yes. good point.
i changed this yesterday when trying to make sense why the tests were failing on the server. but now knowing that the problem is not the python version, we should indeed revert.

PR #117, opened 2024-02-14
Word schedule changes

tfnribeiro:
My proposed changes:

1. Created a new method: `priority_bookmarks_to_study` which uses the priority we have discussed of taking words closest to being learned followed by the lowest `word_rank` in the DB. This needs to query all the current scheduled bookmarks so the ordering is correct.
2. Since similar words would find words that are scheduled for learning, I have kept the original `bookmarks_to_study` though now it doesn't randomize them.
3. So, I checked the [Leitner System Wikipedia](https://en.wikipedia.org/wiki/Leitner_system) article and the previous implementation would follow the original system: If a user gets a bookmark wrong, we start from scratch. Now, the way it works is by using the alternative also shown in the article, where the wrong bookmarks are moved one scheduling backwards. They would still need to recall it 4 times correctly, to "learn" it.

We have 5 levels **0** -> **1** -> **2** -> **4** -> **8** day intervals.

5. Created a method to retrieve the bookmarks for the day: `get_current_study_window`. This method retrieves all the bookmarks that are scheduled before the tomorrow's date at midnight (essentially, all the words that are scheduled in the current date or below).

Let me know what you think!

Below are the images from the Wikipedia article visualizing the differences between the standard Leitner and the alternative that is now implemented.

## Old Scheduling:
image

## New Scheduling:
image


zeeguu/core/model/user.py
```diff
@@ -201,21 +201,35 @@ def has_bookmarks(self):
...
+            to_study = BasicSRSchedule.priority_bookmarks_to_study(self, bookmark_count)
 
         return to_study
+    
+    def scheduled_bookmarks(self, bookmark_count=10):
+        """
+        :param bookmark_count: by default we recommend 10 words
+        :return: a list of 10 words that are scheduled to be learned.
+        """
+        db_session = zeeguu.core.model.db.session
```
mircealungu, line 228: this db_session is never used; can probably remove the line?

zeeguu/core/model/user.py
```diff
@@ -201,21 +201,35 @@ def has_bookmarks(self):
 
     def bookmarks_to_study(self, bookmark_count=10):
         """
+        We now use a logic to sort the words, if we call this everytime
```
mircealungu, line 204: Uff. It's hard for me to make sense of the difference between these two. 
What I wonder now is the following:
- the similar words is only used in MultipleChoice.js

Why don't we ask for more words at the beginning of the exercise session, and just pass them to the exercise at creation time. Then we don't have to reach out from within the exercise to get "similar words". 

We can then completely remove the similar words function and endpoint, which is anyway not a great implementation...

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -52,29 +58,36 @@ def update_schedule(self, db_session, correctness):
...
-            next_practice_date = datetime.now() + timedelta(
-                minutes=new_cooling_interval
+            # Since we can now loose the streak on day 8,
+            # we might have to repeat it a few times to learn it.
+            new_cooling_interval = NEXT_COOLING_INTERVAL_ON_SUCCESS.get(
+                self.cooling_interval, 8 * ONE_DAY
             )
-            self.cooling_interval = new_cooling_interval
-            self.next_practice_time = next_practice_date
             self.consecutive_correct_answers += 1
```
mircealungu, line 76: i guess with the leitner system we don't even need the consecutive_correct_answers?

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -52,29 +58,36 @@ def update_schedule(self, db_session, correctness):
...
-
-            new_cooling_interval = NEXT_COOLING_INTERVAL_ON_SUCCESS[
-                self.cooling_interval
-            ]
-            next_practice_date = datetime.now() + timedelta(
-                minutes=new_cooling_interval
+            # Since we can now loose the streak on day 8,
+            # we might have to repeat it a few times to learn it.
+            new_cooling_interval = NEXT_COOLING_INTERVAL_ON_SUCCESS.get(
+                self.cooling_interval, 8 * ONE_DAY
```
mircealungu, line 74: It's not readily clear to me why the `.get.(..., 8 days)` pattern? i guess if we're on day 8 we get another day 8 interval?

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -42,7 +48,7 @@ def __init__(self, bookmark=None, bookmark_id=None):
 
     def update_schedule(self, db_session, correctness):
         if correctness:
-
+            # Why have CORRECTS_IN_A_ROW_FOR_LEARNED = 4 rather than 3?
```
mircealungu, line 51: asked also below: do we still need corrects in a row with Leitner ?

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -95,6 +108,18 @@ def update(cls, db_session, bookmark, outcome):
         schedule = cls.find_or_create(db_session, bookmark)
         schedule.update_schedule(db_session, correctness)
 
+    @classmethod
+    def get_current_study_window(cls):
```
mircealungu, line 112: cool.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +141,50 @@ def find_or_create(cls, db_session, bookmark):
...
+    def priority_bookmarks_to_study(cls, user, required_count):
+        """
+        Prioritizes the bookmarks to study. To randomize the
+        exercise order utilize the Frontend assignBookmarksToExercises.js
+
+        The original logic is kept in bookmarks_to_study as it is called to
+        get similar_words to function as distractors in the exercises.
+
+        Currently, we prioritize bookmarks in the following way:
+        1. Words that are closest to being learned (indicated by `consecutive_correct_answers`)
```
mircealungu, line 154: ibidem.

mircealungu (approved):
Nice work Tiago.
I have two main questions:
- do we really need to track consecutive corrects? (isn't the next_scheduled_interval) sufficient as a measure of progress?
- should we remove completely the `similar_words`?

See the details in the comments above.

zeeguu/core/model/user.py
tfnribeiro: That could be a way to go. 

I also did work on the similar words in my own project using those confusion word sets and the NLP similarity, so maybe at some point we can just move the similar word call to query that rather than the current logic. However, I think it could be nice to have a way of `get_user_words(n)` to get some words the user has seen in the system. 

The reason why I left it in the system is because I believe it might still be useful as a way to "distract" users but ensure it's words they have seen and not something random.

So maybe we should just rename the end point? What do you think?

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: True, however we need to make sure we update the strings to reflect that we do not care about the consecutive number of answers.

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: The reason was because if we keep track of the consecutive answers, I would then just schedule the word every 8 days until the user gets it correct. But I will remove the consecutive number of answers.

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: I will leave the code to keep track of the consecutive answers, but not use it in the logic and instead using the cooling interval. 

Should we update the strings in the Web explaining the "Learned" words? "Learned words are words that were correct in exercises on four different days or words that have been tagged to be 'Too easy'." Or is it OK still?

tfnribeiro:
I agree with not using the consecutive words. I left the code to still track it, but now the cooldown is checked.

`if self.cooling_interval == MAX_INTERVAL_8_DAY`

I think the similar words we could at some point use the NLP models to more realistically get "similar" words, but I would keep it, but change to a more descriptive name, maybe like `get_user_bookmarks`?

merleschoen (approved):

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: Not necessarily. You could have a sequence like: C (1) -> C (2) -> C (3) -> W(2) -> C(3) -> C(4). I think we could just say "at least 4 different days" and that would fix the technicality.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+    def priority_bookmarks_to_study(cls, user, required_count):
+        """
+        Prioritizes the bookmarks to study. To randomize the
+        exercise order utilize the Frontend assignBookmarksToExercises.js
+
+        The original logic is kept in bookmarks_to_study as it is called to
+        get similar_words to function as distractors in the exercises.
+
+        Currently, we prioritize bookmarks in the following way:
+        1. Words that are closest to being learned (indicated by `cooling_interval`, the highest the closest it is)
```
mircealungu, line 156: Cool!

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+            cooling_interval = (
+                cls.query.filter_by(bookmark=bookmark).one().cooling_interval
+            )
+            user_word = UserWord.query.filter_by(id=bookmark.origin_id).one()
+            word_rank = user_word.rank
+            if word_rank is None:
+                word_rank = 10000
+            return (-cooling_interval, word_rank)
+
+        tomorrow = cls.get_current_study_window()
```
mircealungu, line 170: `end_of_today` is slightly more precise than tomorrow? 
Same question elsewhere where `tomorrow` is being used.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+
+        def sorting_properties(bookmark):
+            cooling_interval = (
+                cls.query.filter_by(bookmark=bookmark).one().cooling_interval
+            )
+            user_word = UserWord.query.filter_by(id=bookmark.origin_id).one()
+            word_rank = user_word.rank
+            if word_rank is None:
+                word_rank = 10000
+            return (-cooling_interval, word_rank)
```
mircealungu, line 168: my IDEs suggestion: image

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+            Bookmark.query.join(cls)
+            .filter(Bookmark.user_id == user.id)
+            .join(UserWord, Bookmark.origin_id == UserWord.id)
+            .filter(UserWord.language_id == user.learned_language_id)
+            .filter(cls.next_practice_time < tomorrow)
+            .all()
+        )
+
+        # Remove possible duplicated words from the list
+        # - The user might have multiple repeated words.
```
mircealungu, line 183: Maybe: 

> The user might have multiple *translations of the same word in different contexts that are saved as different bookmarks*

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+        # - In a session, it should just show up once.
```
mircealungu, line 184: Maybe: 

> ... *a word* should only show up once

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+
+        The original logic is kept in bookmarks_to_study as it is called to
+        get similar_words to function as distractors in the exercises.
+
+        Currently, we prioritize bookmarks in the following way:
+        1. Words that are closest to being learned (indicated by `cooling_interval`, the highest the closest it is)
+        2. Words that are most common in the language (utilizing the word rank in the db)
+        """
+
+        def sorting_properties(bookmark):
```
mircealungu, line 160: PyCharm suggestion: 

image

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+            cooling_interval = (
+                cls.query.filter_by(bookmark=bookmark).one().cooling_interval
```
mircealungu, line 162: But, we already have the bookmark object loaded in memory. We should be able to simply say: 

```
bookmark.cooling_interval
```

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -116,9 +143,63 @@ def find_or_create(cls, db_session, bookmark):
...
+            )
+            user_word = UserWord.query.filter_by(id=bookmark.origin_id).one()
+            word_rank = user_word.rank
+            if word_rank is None:
+                word_rank = 10000
```
mircealungu, line 167: We already have a constant called `UserWord.IMPOSSIBLE_RANK` - should we not use that? 
We might want to rename that, and think about it's value. 
- rename to: `MAX_RANK`? `UNKNOWN_RANK`? keep `IMPOSSIBLE_RANK`? `NO_RANK`? 
- value: 
    1. start using 1,000,000 everywhere? that is clearly future proof, if at some point we want to add more ranks to the DB than 10,000. 
    2. set that constant to 10,000 - we don't have to update the DB to make all the 10,000 into 1MIL... that's not an argument, it's a simple update.

zeeguu/core/model/user.py
mircealungu: You're right. Distraction words that come from the set of translated words still makes sense. How would you rename the endpoint? Something like: `get_random_bookmarks`? Could take as input a set of bookmark ids to be excluded, so we don't by chance stumble upon words that are being studied at the moment?

mircealungu (changes requested):
overall cool stuff!

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: I am happy to remove the parenthesis, I only use them to make it clear it returns a tuple. I don't know what you think.

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: I am not sure what the comment of "Shadows name", is it because we have a class called Bookmark?

zeeguu/core/model/user.py
tfnribeiro: Mhmm, the thing is that it's not random, it's from the scheduled_list, maybe just calling filtered_scheduled_bookmarks(id_set) would make the most sense, so we could pass it the ids of the scheduled words and avoid collisions, and if not possible we resort to similar words?

zeeguu/core/word_scheduling/basicSR/basicSR.py
tfnribeiro: I tried this, but I get an error: 

2024-02-21 17:12:15   File "/Zeeguu-API/zeeguu/core/word_scheduling/basicSR/basicSR.py", line 161, in sorting_properties
2024-02-21 17:12:15     cooling_interval = bookmark.cooling_interval
2024-02-21 17:12:15 AttributeError: 'Bookmark' object has no attribute 'cooling_interval'

I don't think it actually stores it from the query, or should I do something else?

mircealungu: sorry, this is stupid. Indeed, the cooling interval is in the BasicSR object, not in the bookmark.

zeeguu/core/model/user.py
tfnribeiro: It could, but again we are sorting the objects every time so I don't think it would be smart to do that. Instead, we would have a more "relaxed" query that just excludes the ids and doesn't need to sort them or filter them by duplicates.

mircealungu: you could take one extra parameter for sorting? :) 
that being said, that method would become more complicated maybe? in that case let's go with your idea!

mircealungu:
let's merge then and we leave the method renames for another time!

PR #124, opened 2024-03-01
Move is dev flag to user table

tfnribeiro:
Not all devs necessarily will have a Teacher account and vice-versa. Instead, the information will be stored in the user table.


mircealungu:
Ah, btw, let's try to keep a bit track of the date of the migrations in their file-name. As in this commit: 

[image
](https://github.com/zeeguu/api/commit/880b1065abe2689b7befb4a2d3a2e2ae3ad48b64)

PR #126, opened 2024-03-06
Proposed changes

oskaitu:
Proposed changes to API:
- Changed liked column in database to be set to None when created. Adding the functionality to see explicit dislike.
- Removing EVENT_LIKE_ARTICLE from distill_article_interactions. This function would make the liked column be set to True no matter the user input from the frontend(If they pressed yes or no). because logUserActivity from the frontend, sent an EVENT_LIKE to the backend which was caught by distill_article_interactions.

### DIFF
--- a/zeeguu/core/user_activity_hooks/article_interaction_hooks.py
+++ b/zeeguu/core/user_activity_hooks/article_interaction_hooks.py
@@ -28,10 +28,10 @@ def distill_article_interactions(session, user, data):
 
     if EVENT_OPEN_ARTICLE in event:
         article_opened(session, article_id, user)
-    elif EVENT_LIKE_ARTICLE in event:
-        article_liked(session, article_id, user, True)
     elif EVENT_USER_FEEDBACK in event:
         article_feedback(session, article_id, user, value)
+    """ elif EVENT_LIKE_ARTICLE in event:
+        article_liked(session, article_id, user, True) """
 
 
 def article_feedback(session, article_id, user, event_value):
--- a/zeeguu/core/model/user_article.py
+++ b/zeeguu/core/model/user_article.py
@@ -132,7 +132,7 @@ class UserArticle(db.Model):
         user: User,
         article: Article,
         opened=None,
-        liked=False,
+        liked=None,
         starred=None,
     ):
         """


zeeguu/core/user_activity_hooks/article_interaction_hooks.py
```diff
@@ -28,10 +28,10 @@ def distill_article_interactions(session, user, data):
 
     if EVENT_OPEN_ARTICLE in event:
         article_opened(session, article_id, user)
-    elif EVENT_LIKE_ARTICLE in event:
-        article_liked(session, article_id, user, True)
     elif EVENT_USER_FEEDBACK in event:
         article_feedback(session, article_id, user, value)
+    """ elif EVENT_LIKE_ARTICLE in event:
```
tfnribeiro, line 33: So currently, the API was only checking if the event was "LIKED ARTICLE" and if so, then it would set the article to true. To me it seems right now the button doesn't trigger anything in the API.

One possible solution is:
```
elif EVENT_LIKE_ARTICLE in event:
        article_liked(session, article_id, user, value == "true")
```
Then it will detect the like event and update the value to the user. Doing this seems to correctly store the information in my environment.

There is a question if we should instead add a different event EVENT_DISLIKE_ARTICLE and then we set it to false. Tagging @mircealungu .

oskaitu: Just tried it out, adding the equality check does fix the initial problem that it would just add True.

But in this case, as I understand it, it looks like it's updating the value twice. As we can see in the LikeFeedbackBox, there's a function that updates the backend with the liked state:
LikeFeedbackBox.js:
```javascript
  function setLikedState(state) {
    let newArticleInfo = { ...articleInfo, liked: state };
    api.setArticleInfo(newArticleInfo, () => {
      setArticleInfo(newArticleInfo);
    });
    api.logReaderActivity(api.LIKE_ARTICLE, articleID, state, UMR_SOURCE);
  }
```
userArticles.js
```javascript
Zeeguu_API.prototype.setArticleInfo = function (articleInfo, callback) {
  this._post(
    `user_article`,
    `article_id=${articleInfo.id}` +
      `&starred=${articleInfo.starred}` +
      `&liked=${articleInfo.liked}`,
    callback
  );
};
```
This post request is sent to the API and updates the values in this function:
user_article.py:
```python
# ---------------------------------------------------------------------------
@api.route("/user_article", methods=("POST",))
# ---------------------------------------------------------------------------
@cross_domain
@with_session
def user_article_update():
    """

        update info about this (user x article) pair
        in the form data you can provide
        - liked=True|1|False|0
        - starred -ibidem-

    :return: json as prepared by content_recommender.mysql_recommender.user_article_info

    """

    article_id = int(request.form.get("article_id"))
    starred = request.form.get("starred")
    liked = request.form.get("liked")

    article = Article.query.filter_by(id=article_id).one()

    user_article = UserArticle.find_or_create(db_session, flask.g.user, article)

    if starred is not None:
        user_article.set_starred(starred in ["true", "True", "1"])

    if liked is not None:
        user_article.set_liked(liked in ["true", "True", "1"])

    db_session.commit()

    return "OK"
```

This is what we thought was the intended way to do it.

But I guess just using 'api.logReaderActivity(api.LIKE_ARTICLE, articleID, state, UMR_SOURCE);' would be sufficient now as the equality check has been added. In this case, another event for dislike would be easy to add :)

mircealungu: @oskaitu - good observation about writing the info twice. 

I think the `distill_article_interactions` pattern was there before we had the `setArticleInfo` was added as an endpoint. 

I think we could go both ways:
1. remove the `distill` pattern and explicitly call `setArticleInfo` 
    - more readable code - it's clear what's happening from the front-end
    - two calls to the API if we also want to log
3. do not call the `api.setArticleInfo` anymore for like state update
    - a single call to the API => faster
    - the side effect of writing to another table when adding an user activity data event might not be obvious

When choosing between clarity and speed, at this point, I'd go with clarity...

tfnribeiro: That's a good point, I didn't see the session update. That being said, then I don't have any issues (using your PRs in the frontend + backend) the articles update correctly. 

See here:

image
image

image
image

image

Could you guys double check that it doesn't update it ?

oskaitu: Option 1. does seem like the best way to do it, we were at least a bit confused that the logger was updating information in the database. 
The event of OPEN_ARTICLE and LIKE_ARTICLE looks like they already has their own API call to the backend. But it looks like the logger is still being used to update the backend with the EVENT_USER_FEEDBACK.
ReportBroken.js:
```javascript
  function reportBroken() {
          api.logReaderActivity(api.USER_FEEDBACK, articleID, feedback, UMR_SOURCE);
          setIsFeedbackSent(true);
          setTimeout(() => handleClose(), 1000);
  }
```
Is it safe to assume that we can just remove the calls for liked_article and open_article in the distill function, and keep the user_feedback where it is?

```python
def distill_article_interactions(session, user, data):
    """

        extracts info from user_activity_data

    :param session:
    :param event:
    :param value:
    :param user:
    """

    event = data["event"]
    value = data["value"]
    article_id = int(data["article_id"])

    log(f"event is: {event}")

    if EVENT_OPEN_ARTICLE in event:
        article_opened(session, article_id, user)
    elif EVENT_LIKE_ARTICLE in event:
        article_liked(session, article_id, user, value == "true")
    elif EVENT_USER_FEEDBACK in event:
        article_feedback(session, article_id, user, value)
```

oskaitu:
I have double-checked that clicking yes or no in the LikeFeedbackBox updates user_activity_data and user_article in the database as intended. I have commented out the EVENT_LIKE_ARTICLE (for now) as it doesn't have any effect on the outcome.

PR #131, opened 2024-03-26
Allow DB Feeds to skip Metadata retrieval

tfnribeiro:
Added a flag to the FeedHandler super class that allows skipping the extracting metadata call. This should only be called when the  feed is created, which at this point is done in feed.py line 81 `feed_handler = FEED_TYPE_TO_FEED_HANDLER[feed_type](url, feed_type)`. 

When crawling the websites to extract the data, this information is stored in the DB and therefor we do not need to re-acquire it. This is done during `initializeFeedHandler` where now the keyed argument **is_stored_db** is passed to skip the call to `extract_feed_metadata`.

+ Automatic formatting.

### Tests performed:

- Add a new feed and ensure the fields are updated. 
- Crawled the websites and checked the metadata wasn't retrieved.

### DIFF
--- a/zeeguu/core/feed_handler/feed_handler.py
+++ b/zeeguu/core/feed_handler/feed_handler.py
@@ -3,15 +3,12 @@ from zeeguu.core.util.time import normalize_to_server_time
 
 
 class FeedHandler:
-    def __init__(self, url: str, feed_type: int, is_stored_db: bool = False):
+    def __init__(self, url: str, feed_type: int):
         self.url = url
         self.feed_type = feed_type
         self.title = ""
         self.description = ""
         self.image_url_string = ""
-        if not is_stored_db:
-            print("Extracting Feed Metadata.")
-            self.extract_feed_metadata()
 
     def get_server_time(self, article_date) -> datetime:
         if type(article_date) is datetime:
--- a/zeeguu/core/feed_handler/feed_handler.py
+++ b/zeeguu/core/feed_handler/feed_handler.py
@@ -3,13 +3,15 @@ from zeeguu.core.util.time import normalize_to_server_time
 
 
 class FeedHandler:
-    def __init__(self, url: str, feed_type: int):
+    def __init__(self, url: str, feed_type: int, is_stored_db: bool = False):
         self.url = url
         self.feed_type = feed_type
         self.title = ""
         self.description = ""
         self.image_url_string = ""
-        self.extract_feed_metadata()
+        if not is_stored_db:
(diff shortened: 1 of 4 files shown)


zeeguu/core/feed_handler/rssfeed.py
```diff
@@ -6,8 +6,9 @@
 
 
 class RSSFeed(FeedHandler):
-    def __init__(self, url: str, feed_type: int):
-        super().__init__(url, feed_type)
+    def __init__(self, url: str, feed_type: int, is_stored_db: bool = False):
+        super().__init__(url, feed_type, is_stored_db)
+        logp(f"Created RSS Source ({self.url})")
```
mircealungu, line 11: I removed this logline because it was confusing. Given that our feeds are mapped on to the DB, the "created" message should only appear when a new object is added to the DB. Or is this the case here?

zeeguu/core/feed_handler/feed_handler.py
```diff
@@ -3,13 +3,15 @@
...
 class FeedHandler:
-    def __init__(self, url: str, feed_type: int):
+    def __init__(self, url: str, feed_type: int, is_stored_db: bool = False):
         self.url = url
         self.feed_type = feed_type
         self.title = ""
         self.description = ""
         self.image_url_string = ""
-        self.extract_feed_metadata()
+        if not is_stored_db:
```
mircealungu, line 12: There is something strange about this pattern. 
Could we leave the constructor to be just simply initializing the attributes with their default values, and then for those parts of the code where you would call `__init__` with `is_stored_in_db=True` call the `new_feed.extract_feed_metadata()`?

mircealungu (commented):
a small question for you.

zeeguu/core/feed_handler/feed_handler.py
tfnribeiro: I like that! Sure, we can do that!

zeeguu/core/feed_handler/rssfeed.py
tfnribeiro: True, that isn't necessarily a creating a source, but rather using the RSS handler, so I will rephrase it to "Using RSS Handler" and "Using Newspaper Handler" what do you think ?

tfnribeiro:
I have applied the suggested changes - let me know what you think!

zeeguu/core/feed_handler/feed_handler.py
mircealungu: Tak!

PR #132, opened 2024-04-03
updated dependencies in requirements.txt and githooks

merleschoen:


mircealungu (changes requested):
cool. thanks @merleschoen !

@tfnribeiro - we've had to update sci-py version and sci-kit versions with Merle, because  the pinned versions were too old for the Python 3.12.2 on her machine. 

the major version didn't change for neither of them, so conformting to semver the new versions should be completely backward compatible. 

@merleschoen - i remember that we did install some things from the terminal directly with pip on your machine. do you remember what they were? I think `cryptography` and something about lxml... to remember which of the lxml packages we installed try to run `pip freeze` in your venv: that should show you all the packages that you have installed locally. lxml_html_parse or what was the name should be in there and we should also add it to the requirements.txt as a pre-requisite.

tfnribeiro:
If we updated the scikit-learn library then I will have to update the ML model as scikit-learn is very clear that we should not load models with different versions from where they were trained. It shouldn't take too long, so I will push it to this PR if we decide to update Python + these Libraries

mircealungu:
Oh, thanks for pointing this out @tfnribeiro! I wasn't aware of this! In fact, I think it was only the scipy that was complaining, and the sci-kit learn we updated while trying to fix the issue - but might not even be necessary. 

We can go both ways from here: 
1. @merleschoen - could try to downgrade scikit-learn to  `siikit-learn==1.4.0` as it was before, and then run `pip install -r requirements.txt` to see whether it still works; if it does we can just push this for now
2. @tfnribeiro updates the model - although it seems a bit of a waste of time because I don't expect there's going to be much difference betwen sklearn 1.4.0 and 1.4.1.

So I'd propose we try the solutions in the order in which I presented them above!

tfnribeiro:
Sounds good! I will wait until there is a feedback from Merle. I think I was even able to run it with a higher version of 1.5.* of scikit-learn. However, when I read about the persistence of the models, they seemed to advise against it.

I will put their documentation and quote for future reference: 

https://scikit-learn.org/stable/model_persistence.html

> While models saved using one version of scikit-learn might load in other versions, this is entirely unsupported and inadvisable. It should also be kept in mind that operations performed on such data could give different and unexpected results.

merleschoen:
It does work with the downgraded version, so I've updated the requirements.txt accordingly. I've also added the packages we've installed locally.

mircealungu (approved):
tak @merleschoen !

PR #133, opened 2024-04-03
Fix ElasticSearch Topic Search

tfnribeiro:
We were expecting a Keyword to add user topics to the filter the user topics selected.


mircealungu (approved):
that was a nasty bug! tak @tfnribeiro

PR #134, opened 2024-04-18
Learning cycle schedule implementation

merleschoen:
This PR contains the learning cycle column in the bookmark table, settings for disabling productive exercises and an updated scheduling algorithm.


.githooks/pre-commit
tfnribeiro, line 1: Are these changes needed?

zeeguu/core/model/bookmark.py
```diff
@@ -5,6 +5,7 @@
 from sqlalchemy.orm import relationship
 from sqlalchemy.orm.exc import NoResultFound
 from wordstats import Word
+from sqlalchemy import Enum
```
tfnribeiro, line 8: This import is not used anymore (?)

zeeguu/core/model/bookmark.py
```diff
@@ -197,6 +204,18 @@ def json_serializable_dict(self, with_context=True, with_title=False):
...
+        from zeeguu.core.model import BasicSRSchedule
+        from zeeguu.core.word_scheduling.basicSR.basicSR import MAX_INTERVAL_8_DAY, ONE_DAY
+
+        try:
+            basic_sr_schedule = BasicSRSchedule.query.filter(
+                BasicSRSchedule.bookmark_id == self.id
+            ).one()
+            cooling_interval = basic_sr_schedule.cooling_interval // ONE_DAY
+        except sqlalchemy.exc.NoResultFound:
+            cooling_interval = -1
```
tfnribeiro, line 217: I know I was the one to set this to -1, but it might be better to set it to a None so it's clear that it's not set or doesn't exist. I think it better future proofs in case some implementation would have a negative scale.

tfnribeiro (changes requested):
Overall really good, just a few questions and a suggestion to make it more future proof.

zeeguu/api/endpoints/feature_toggles.py
```diff
@@ -51,6 +52,13 @@ def _tiago_exercises(user):
     return right_user and right_language
 
 
+def _merle_exercises(user):
+    # Change this to Match Merle's Exercises
+    # 534: Mircea, 4089: Tiago Test User, 4192: Merle's User
```
mircealungu, line 57: I guess if we don't put this comment on GH, then there is a lower chance that somebody can identify our data in the anonymized DB :)

zeeguu/api/endpoints/translation.py
```diff
@@ -89,6 +91,7 @@ def get_one_translation(from_lang_code, to_lang_code):
             to_lang_code,
             context,
             article_id,
+            learning_cycle=LearningCycle.RECEPTIVE if _merle_exercises(flask.g.user) else LearningCycle.NOT_SET,
```
mircealungu, line 94: The problem here is: what is the meaning of this column? Learning cycle? Is it the current learning cycle? If it has not been scheduled yet, then I think it's unfair to say that it's in either of the cycles. 

Thus, what I would propose is that here, you initialize *all* of the Bookmarks with `NOT_SET` and then only the first time a word is seen in an exercise, this column gets to be set. Do you see what I mean?

zeeguu/core/model/bookmark.py
```diff
@@ -67,23 +69,28 @@ class Bookmark(db.Model):
...
     ):
         self.origin = origin
         self.translation = translation
         self.user = user
         self.time = time
         self.text = text
-        self.stared = False
+        self.starred = False
         self.fit_for_study = fit_for_study(self)
+        self.learned = False
```
mircealungu, line 92: nice spotting!

zeeguu/core/model/bookmark.py
mircealungu, line 217: argh! i clicked Resolve on Tiago's comment, instead of reply. I agree with him. Setting this to None would be slightly more intutive.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -46,19 +48,40 @@ def __init__(self, bookmark=None, bookmark_id=None):
...
             if self.cooling_interval == MAX_INTERVAL_8_DAY:
-                self.bookmark.learned = True
-                self.bookmark.learned_time = datetime.now()
-                db.session.add(self.bookmark)
-                db.session.commit()
-                db.session.delete(self)
-                db.session.commit()
-                return
+                if learning_cycle == LearningCycle.NOT_SET:
+                    self.set_bookmark_as_learned(db_session)
```
mircealungu, line 69: I think here you can reverse the conditions maybe for simpler code? I.e. if `learningCycle == RECEPTIVE and setting=="true"` then do the switch of the learning cycle, otherwise for all the other cases we seem to be doing the "`set_bookmark_as_learned(db_session)`". Or?

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -169,16 +193,24 @@ def sorting_properties(bookmark):
...
             .filter(Bookmark.user_id == user.id)
             .join(UserWord, Bookmark.origin_id == UserWord.id)
             .filter(UserWord.language_id == user.learned_language_id)
             .filter(cls.next_practice_time < end_of_day)
-            .all()
         )
 
+         # If productive exercises are disabled, exclude bookmarks with learning_cycle of 2
+        if productive_exercises_setting == "false":
+            scheduled_candidates_query = scheduled_candidates_query.filter(Bookmark.learning_cycle == 1)
```
mircealungu, line 210: can we use the constant here? Would cleaner and easier to read :)

mircealungu (changes requested):

zeeguu/api/endpoints/translation.py
mircealungu: Alternatively, let's leave it like this for now. Then when @tfnribeiro merges his own changes that pertain the exercises, he makes this change himself. He already has written code for detecting that moment when a bookmark is entering the "studying pipeline".

mircealungu:
I'm getting a bit of a problem here. 

I suspect that my the call to `.first()` returns None. And thus we can't do value. 
What I think we need is a default fallback: if the user does not have a preference set, we use the default one. The question is: do we save a default one in the DB? Or we just assume that a normal learner would by default expect productive exercises, so if there's no preference in the DB, we return True here? 

```python
  File "/Users/gh/zeeguu/api/zeeguu/core/word_scheduling/basicSR/basicSR.py", line 197, in priority_bookmarks_to_study
    productive_exercises_setting = UserPreference.query.filter_by(user_id=user.id, key="productive_exercises").first().value
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'value'
```

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -46,19 +48,41 @@ def __init__(self, bookmark=None, bookmark_id=None):
...
-            # the word.
             if self.cooling_interval == MAX_INTERVAL_8_DAY:
-                self.bookmark.learned = True
-                self.bookmark.learned_time = datetime.now()
-                db.session.add(self.bookmark)
-                db.session.commit()
-                db.session.delete(self)
-                db.session.commit()
-                return
+                if learning_cycle == LearningCycle.NOT_SET:
```
tfnribeiro, line 68: This logic won't be quite the same after changing the way we progress the words correct? 
Now, NOT_SET will only be used when words are not schedule.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -232,6 +265,8 @@ def schedule_some_more_bookmarks(cls, session, user, required_count):
             id = b["bookmark_id"]
             b = Bookmark.find(id)
             print(f"scheduling another bookmark_id for now: {id} ")
+            b.learning_cycle = LearningCycle.RECEPTIVE.value
```
tfnribeiro, line 268: Since I got errors without using `.value`, we might want to use this syntax for all enums, where we are using Enums. I remember we discussed it might be due to the new version of python? Maybe @mircealungu can check if he gets any errors.

tfnribeiro (changes requested):
Just some small notes!

tools/migrations/24-05-update_learning_cycle_preference_for_all_users.sql
```diff
@@ -0,0 +1,7 @@
+/* 
+ Updates the learning cycle preference enabling productive exercises for all users.
+ */
+INSERT INTO user_preference (user_id, `key`, value)
+SELECT u.id, 'productive_exercises', 'true'
+FROM `user` AS u
+WHERE u.id NOT IN (SELECT user_id FROM user_preference WHERE `key` = 'productive_exercises');
```
mircealungu, line 7: looks cool!

tools/migrations/24-05-update_learning_cycle_for_existing_users.sql
```diff
@@ -0,0 +1,6 @@
+/*
```
mircealungu, line 1: let's try to use the following naming convention from today on: https://github.com/zeeguu/api/blob/master/tools/migrations/README.md#convention-for-naming-migrations

zeeguu/core/model/user.py
```diff
@@ -727,11 +727,6 @@ def authorize_anonymous(cls, uuid, password):
     def create_default_user_preference(self):
         from zeeguu.core.model.user_preference import UserPreference
 
-        existing_preference = UserPreference.find_or_create(
-            db.session, self, UserPreference.PRODUCTIVE_EXERCISES
+        UserPreference.find_or_create(
+            db.session, self, UserPreference.PRODUCTIVE_EXERCISES, "true"
```
mircealungu, line 731: we simplified this a bit - if you pass the value, you already have the code initializing the preference accordingly

zeeguu/api/endpoints/translation.py
mircealungu, line 94: Ok, so this should be now initialized to NOT_SET after the discussions of yesterday. right?

zeeguu/api/endpoints/user_preferences.py
```diff
@@ -35,6 +35,14 @@ def save_user_preferences():
         pref.value = audio_exercises_value
         zeeguu.core.model.db.session.add(pref)
 
+    productive_exercises_value = data.get(UserPreference.PRODUCTIVE_EXERCISES, None)
+    if productive_exercises_value:
+        pref_productive = UserPreference.find_or_create(
+            zeeguu.core.model.db.session, flask.g.user, UserPreference.PRODUCTIVE_EXERCISES
+        )
+        pref_productive.value = productive_exercises_value
```
mircealungu, line 43: good!

zeeguu/api/endpoints/translation.py
mircealungu: Or just allow the default parameter value from the Bookmark constructor to take care of this. What I mean is that if you don't pass this, it's going to be set to NOT_SET anyway.

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -47,18 +50,31 @@ def __init__(self, bookmark=None, bookmark_id=None):
...
-            # the word.
             if self.cooling_interval == MAX_INTERVAL_8_DAY:
-                self.bookmark.learned = True
-                self.bookmark.learned_time = datetime.now()
-                db.session.add(self.bookmark)
-                db.session.commit()
-                db.session.delete(self)
-                db.session.commit()
-                return
+                if learning_cycle == LearningCycle.RECEPTIVE & productive_exercises_enabled:
```
mircealungu, line 68: should be `booleand and boolean` not `boolean & boolean`

zeeguu/core/word_scheduling/basicSR/basicSR.py
```diff
@@ -232,6 +255,8 @@ def schedule_some_more_bookmarks(cls, session, user, required_count):
             id = b["bookmark_id"]
             b = Bookmark.find(id)
             print(f"scheduling another bookmark_id for now: {id} ")
+            b.learning_cycle = int(LearningCycle.RECEPTIVE)
```
mircealungu, line 258: no need for this casting.

mircealungu (changes requested):
very minor changes, and this is ready to be merged and deployed!

tfnribeiro (approved):
Reviewed together with Mircea.

PR #138, opened 2024-04-29
Update user_preference.py

tfnribeiro:
+ Fix to access the value rather than the UserPreference object.


mircealungu (approved):
tak @tfnribeiro

PR #140, opened 2024-04-30
Exercise notification new exercises

tfnribeiro:
I have decided to add this PR which introduces some new endpoints for the exercises.

This will work in tandem with the new way we have discussed to introduce exercises to the users. 

**Main Idea:** Users are now introduced to the words until a limit (50) - this is defined at the frontend as a constant, at which point the system will alert them that they already have this number of words in learning. The user can overrule this limit, by asking to start learning more words. 

The system also checks if the user has any scheduled words to review when they log-in to the app. These are given first to the user so he can progress in the schedule, and only after are new words introduced to start learning. These words will be marked as new with a small notification icon.

**Note:** We also revert the preference of users to having default exercises by default, these instead should be only introduced to new users, when the exercises are introduced. Otherwise, the user would have to complete 2 cycles before reaching the learned status.

_Some formatting changes, all done using black-formatter_


mircealungu:
One thing we observed with @merleschoen yesterday is that currently, a user can choose to do exercises after they read a text. The implementation currently has them doing exercises with all the translations in the text. We have to think whether this makes sense anymore at the moment. 

One other thing to think about: do we cap the _total words in learning_ or the _total words that they struggle with_? I might be able to take on more if somehow in the words that are scheduled are actually easy for me. Why did I translate them in the first place you ask me... I just wasn't sure - but every time I see them in exercises I actually get them now... This whole thing might still not be needed though: if the user wants to add more words in the pipeline they should be able to go to their Words tab and remove some from practice I guess.

tfnribeiro:
Personally, _total words that they struggle with_ seems a bit hard to estimate or even when looking at exercises without having some history. I think one could try to user other user data, but it can be easy to mislabel. I think if we want to give the user more power, then it should be easy for them to "unschedule" words, which we do via the button "Don't show this word again". 

image

Again, the user still has the power to add more words, and we can set a limit higher than 50 to around 100 words. This could be even used based on how much time a user wants to spend in the app. Lower for less dedicated users, higher with more ambitious users.

mircealungu (approved):
cool stuff. 
this is ready to be merged and deployed.

PR #144, opened 2024-05-06
shows all exercises for test cohort no matter the language

merleschoen:
This pr allows the cohort language to be null, so teachers can see all exercises, regardless of the student's chosen language. It also includes a sql query to update the language of the cohort testing the learning cycle implementation to null.

### DIFF
--- a/tools/migrations/24-05-06--update_cohort_language_to_null_for_testing.sql
+++ b/tools/migrations/24-05-06--update_cohort_language_to_null_for_testing.sql
@@ -1,0 +1,4 @@
+/* 
+ Sets language id for learning cycle test cohort to null, so we can see all exercise despite the language.
+ */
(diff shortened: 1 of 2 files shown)


mircealungu (approved):
looks good.

tools/migrations/24-05-06--update_cohort_language_to_null_for_testing.sql
```diff
@@ -0,0 +1,4 @@
+/* 
+ Sets language id for learning cycle test cohort to null, so we can see all exercise despite the language.
+ */
+UPDATE cohort SET language_id = NULL WHERE name IN ('Merle', 'MerleITU');
```
mircealungu, line 4: ah, this has to be `invite_code` not `name`. otherwise good!

PR #146, opened 2024-05-09
FEATURE: content based filtering 

nselpriv:
### This Pull request adds the following functionality

1. New articles like function in elastic-recommender that uses user likes to recommend articles like what they have already explicitly liked 
2. New endpoint for /foryou  

**This PR is combined with a PR in zeeguu-web, they should be approved together https://github.com/zeeguu/web/pull/371**

### DIFF
--- a/zeeguu/api/endpoints/user_articles.py
+++ b/zeeguu/api/endpoints/user_articles.py
@@ -1,10 +1,11 @@
 import flask
 
-from zeeguu.core.content_recommender import article_recommendations_for_user, topic_filter_for_user
+from zeeguu.core.content_recommender import article_recommendations_for_user, topic_filter_for_user, content_recommendations
 from zeeguu.core.model import UserArticle, Article, PersonalCopy
 
 from zeeguu.api.utils.route_wrappers import cross_domain, with_session
 from zeeguu.api.utils.json_result import json_result
+from sentry_sdk import capture_exception
 from . import api
 
 from flask import request
@@ -94,3 +95,22 @@ def user_articles_cohort():
     """
 
     return json_result(flask.g.user.cohort_articles_for_user())
+
+# ---------------------------------------------------------------------------
+@api.route("/user_articles/foryou", methods=("GET",))
+# ---------------------------------------------------------------------------
+@cross_domain
+@with_session
+def user_articles_foryou():
+    article_infos = []
+    try:
+        articles = content_recommendations(flask.g.user.id, flask.g.user.learned_language_id)
+        print("Sending CB recommendations")
+    except Exception as e:
+        print(e)
+        capture_exception(e)
+        #Usually no recommendations when the user has not liked any articles
+        articles = []
+    article_infos = [UserArticle.user_article_info(flask.g.user, a) for a in articles]
+    
+    return json_result(article_infos)
--- a/zeeguu/core/content_recommender/elastic_recommender.py
+++ b/zeeguu/core/content_recommender/elastic_recommender.py
@@ -6,7 +6,6 @@
    - topics, language and user subscriptions.
 
 """
-
 from elasticsearch import Elasticsearch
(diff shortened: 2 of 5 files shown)


zeeguu/core/content_recommender/elastic_recommender.py
```diff
@@ -303,3 +302,78 @@ def _difficuty_level_bounds(level):
...
+    es = Elasticsearch(ES_CONN_STRING)
+    fields = ["content", "title"]
+    language = Language.find_by_id(language_id)
+    like_documents = [
+        {"_index": ES_ZINDEX, "_id": str(doc_id) } for doc_id in recommended_articles_ids
+    ]
+
+    cutoff_date = datetime.now() - timedelta(days=article_age)
+
+    mlt_query = {
```
tfnribeiro, line 317: I would suggest building this query by using a pattern similar of what is used in `build_elastic_recommender_query` under `elastic_query_builder.py`.

nselpriv: @tfnribeiro 
Implemented your suggestion 👍

zeeguu/core/elastic/elastic_query_builder.py
```diff
@@ -182,3 +184,64 @@ def build_elastic_search_query(
     query = {"size": count, "query": weighted_query.to_dict()}
 
     return query
+
+def build_elastic_more_like_this_query(
+    language: Language,
+    like_documents: list[dict[str, str]],
+    similar_to: list[str],
+    cutoff_age: int,
```
tfnribeiro, line 192: `cutoff_days`, might be a better name for this one?

zeeguu/api/endpoints/user_articles.py
```diff
@@ -94,3 +94,20 @@ def user_articles_cohort():
...
+# ---------------------------------------------------------------------------
+@cross_domain
+@with_session
+def user_articles_foryou():
+    article_infos = []
+    try:
+        articles = content_recommendations(flask.g.user.id, flask.g.user.learned_language_id)
+        print("Sending CB recommendations")
+    except:
+        #Usually no recommendations when the user has not liked any articles
```
tfnribeiro, line 109: We don't want to silently catch all errors, make sure to at least log the exception if there is an error.

Currently we use Sentry to keep track of exceptions in zeeguu:

```
    except Exception as e:
        from sentry_sdk import capture_exception

        capture_exception(e)
```

tfnribeiro (commented):
Some extra comments

zeeguu/api/endpoints/user_articles.py
nselpriv: It should be ready now

zeeguu/core/elastic/elastic_query_builder.py
```diff
@@ -182,3 +184,64 @@ def build_elastic_search_query(
...
+    like_documents: list[dict[str, str]],
+    similar_to: list[str],
+    cutoff_days: int,
+    scale: str ="10d",
+    offset: str ="4h",
+    decay: float=0.9):
+    
+    cutoff_date = datetime.now() - timedelta(days=cutoff_days)
+
+    query = {
```
mircealungu, line 199: json is not for humans !!! 
in other queries i built little python DSLs that help build up such a monster w/o having to dump the whole string in the code. i'll accept it like this for now, but if you find the time after you've defended, i'd be happy to receive an updated code version that is less a monster string and more a nicely constructed python expression.

mircealungu (approved):
only one comment, but i'll accept for now so we can merge and deploy! i want to start getting recommendations!

zeeguu/core/elastic/elastic_query_builder.py
nselpriv: I'll make sure to take a look at it again and extend the DSL with what I need and then build the same query with it.
Thanks for the feedback

PR #147, opened 2024-05-14
Add delete user endpoint

tfnribeiro:
+ Expose a way to delete users to the API
+ Uses the user current session id to start the deletion process. 

Used by: https://github.com/zeeguu/web/pull/373


zeeguu/api/endpoints/accounts.py
```diff
@@ -13,6 +13,25 @@
 from zeeguu.logging import log
 
 
+@api.route("/delete_user/<session_id>", methods=["POST"])
+@cross_domain
+def remove_user(session_id):
+    from zeeguu.core.account_management.user_account_deletion import delete_user_account
+
+    try:
+        session_id = int(session_id)
```
mircealungu, line 22: why do we pass the session_id in the URL? Can't we use the @with_session decorator?

mircealungu (changes requested):
Tiago: one more observation on this PR

zeeguu/api/endpoints/accounts.py
tfnribeiro: That's a good point -  I will check if it works using the decorator!

tfnribeiro:
I have changed to use the session. I also added a test case for the user deletion. I remember we talked about a possible issue with ordering, but I don't recall the solution? Right now, it is at the end of all the `user_tests`, where it deletes the `annonymous` user.

zeeguu/core/account_management/user_account_deletion.py
```diff
@@ -0,0 +1,99 @@
...
+    User,
+    UserActivityData,
+    Bookmark,
+    UserArticle,
+    UserReadingSession,
+    UserExerciseSession,
+)
+from zeeguu.core.model import Article
+
+tables_to_modify = [
```
mircealungu, line 33: I remember seeing this elsewhere. Ah, yes, in `delete_dev_users`. Would it be possible to remove it from there, and thus, remove the duplication between the two files. I think also the `delete_user_account` method is duplicated with `delete_user`?

mircealungu (changes requested):
small favor to ask :)

PR #149, opened 2024-05-15
Fix a bug where splitting wasn't mirroring Articles

tfnribeiro:
Created issues where a parsed article would have different count of words when it's downloaded in the database. This resulted in some small paywalled articles being downloaded for FR.

### DIFF
--- a/zeeguu/core/content_quality/quality_filter.py
+++ b/zeeguu/core/content_quality/quality_filter.py
@@ -31,7 +31,7 @@ incomplete_suggesting_terminations = "Read More"
 
 LIVE_BLOG_KIND_OF_PATTERNS = [
     "Lees hier het hele verhaal",
-    "Lees hier het hele verhaal"
+    "Lees hier het hele verhaal",
 ]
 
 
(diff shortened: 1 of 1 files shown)


zeeguu/core/content_quality/quality_filter.py
```diff
@@ -46,8 +46,7 @@ def sufficient_quality_html(html):
 
 
 def sufficient_quality_plain_text(text):
-    word_count = len(text.split(" "))
-
+    word_count = len(text.split())
```
mircealungu, line 49: in which way does this make a difference?

tfnribeiro: The two are not equivalent. If no parameter is passed python looks for any consecuntive white spaces and consideres them as a single seperator:

https://stackoverflow.com/questions/62013468/is-there-a-difference-between-split-vs-split
```
>>> "I am    a    friend!".split()
['I', 'am', 'a', 'friend!']
>>> "I am    a    friend!".split(" ")
['I', 'am', '', '', '', 'a', '', '', '', 'friend!']
>>>
```

mircealungu: whoa. didn't think about consecutive spaces. good catch @tfnribeiro !

PR #151, opened 2024-05-21
Hardening user session management

mircealungu:
code works for me. can you test @tfnribeiro ?

### DIFF
--- a/tools/migrations/24-05-21--add_uuid_to_session.sql
+++ b/tools/migrations/24-05-21--add_uuid_to_session.sql
@@ -1,0 +1,6 @@
+ALTER TABLE
+    `session`
+ADD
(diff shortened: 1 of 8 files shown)


tfnribeiro (approved):
Looks good! 

I couldn't login without using the updated Web version - I am suspecting it has to do with the fact the session is now stored in cookie.

Another thing I have noticed is that now with a login we just add a row to the database without cleaning up. I guess this is lower priority but we should have an expiry date just so we don't keep a very large amount of sessions. 

image

mircealungu:
Yes. Let's open an issue about a more thorough handling of sessions. Besides deletion, we should add:
- renewal
- caching if possible - there was a cache before, that would allow us to avoid hitting the DB to get the user id every time - now we do two queries before every endpoint: getting the session and getting the user...

PR #153, opened 2024-05-22
Updated Session Handling

tfnribeiro:
- Updated naming of decorator from @with_session to @has_session
- Instead of the storing a user object, the flask global now carries the user_id, which should then be handled at each endpoint to retrieve the user object. This avoids retrieving the user object every time we check for the session.
- The session has now a timeout which allows the app to bypass the session table. We also check if the session is older than a month and force the users to relog if that's the case.

### DIFF
--- a/zeeguu/api/endpoints/sessions.py
+++ b/zeeguu/api/endpoints/sessions.py
@@ -10,19 +10,16 @@ from . import api, db_session
 DAYS_BEFORE_EXPIRE = 30  # Days
 
 
-def is_session_valid(session_object):
-    print("----------- Validating the Session based on Last use!! -----------")
-    if session_object is None:
-        return False
-    is_session_too_old = (
-        datetime.now() - session_object.last_use
-    ).days > DAYS_BEFORE_EXPIRE
-    if is_session_too_old:
-        print("Session was too old! - logging out!")
-        db_session.delete(session_object)
-        db_session.commit()
-        return False
-    return True
+def is_session_too_old(session_object):
+    return (datetime.now() - session_object.last_use).days > DAYS_BEFORE_EXPIRE
+
+
+def force_user_to_relog(session_object, reason: str = ""):
+    print(
+        f"Session for user '{session_object.user_id}' was terminated. Reason: '{reason}'"
+    )
+    db_session.delete(session_object)
+    db_session.commit()
 
 
 @api.route("/session/<email>", methods=["POST"])
@@ -92,7 +89,10 @@ def validate():
     """
     # TODO: ideally update in parallel with running the decorated method?
     session_object = Session.find(flask.g.session_uuid)
-    if not is_session_valid(session_object):
+    if session_object is None:
+        flask.abort(401)
+    if is_session_too_old(session_object):
+        force_user_to_relog(session_object, "Session was too old.")
         flask.abort(401)
     session_object.update_use_date()
     db_session.add(session_object)
--- a/zeeguu/api/endpoints/sessions.py
+++ b/zeeguu/api/endpoints/sessions.py
@@ -1,10 +1,29 @@
 import flask
+from datetime import datetime
 from flask import request, make_response
 from zeeguu.core.model import Session, User
 from zeeguu.api.utils.abort_handling import make_error
 
-from zeeguu.api.utils.route_wrappers import cross_domain, with_session
+from zeeguu.api.utils.route_wrappers import cross_domain, has_session
 from . import api, db_session
+
+MAX_TIME_SESSION = 30  # Days
+
+
+def validate_session(session_object):
+    print("----------- Validating the Session based on Last use!! -----------")
+    if session_object is None:
+        flask.abort(401)
+    is_session_too_old = (
+        datetime.now() - session_object.last_use
+    ).days > MAX_TIME_SESSION
+    if is_session_too_old:
+        print("Session was too old! - logging out!")
+        db_session.delete(session_object)
+        db_session.commit()
+        flask.abort(401)
+        return False
+    return True
 
 
 @api.route("/session/<email>", methods=["POST"])
@@ -62,7 +81,7 @@ def get_anon_session(uuid):
 
 @api.route("/validate")
 @cross_domain
-@with_session
+@has_session
 def validate():
     """
 
@@ -72,6 +91,12 @@ def validate():
 
     :return:
     """
+    # TODO: ideally update in parallel with running the decorated method?
+    session_object = Session.find(flask.g.session_uuid)
+    validate_session(session_object)
+    session_object.update_use_date()
+    db_session.add(session_object)
+    db_session.commit()
     return "OK"
 
 
@@ -89,7 +114,7 @@ def is_up():
 
 @api.route("/logout_session", methods=["GET"])
 @cross_domain
-@with_session
+@has_session
 def logout():
     """
 
--- a/zeeguu/api/endpoints/sessions.py
+++ b/zeeguu/api/endpoints/sessions.py
@@ -4,24 +4,23 @@ from flask import request, make_response
 from zeeguu.core.model import Session, User
 from zeeguu.api.utils.abort_handling import make_error
 
-from zeeguu.api.utils.route_wrappers import cross_domain, has_session
+from zeeguu.api.utils.route_wrappers import cross_domain, requires_session
 from . import api, db_session
 
-MAX_TIME_SESSION = 30  # Days
+DAYS_BEFORE_EXPIRE = 30  # Days
 
 
-def validate_session(session_object):
+def is_session_valid(session_object):
     print("----------- Validating the Session based on Last use!! -----------")
     if session_object is None:
-        flask.abort(401)
+        return False
     is_session_too_old = (
         datetime.now() - session_object.last_use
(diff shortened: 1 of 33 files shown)


zeeguu/api/utils/route_wrappers.py
```diff
@@ -1,11 +1,16 @@
...
+
+from datetime import datetime, timedelta
 import zeeguu
 
+SESSION_CACHE = {}
+SESSION_CACHE_EXPIRE = 60  # Seconds
+
 
-def with_session(view):
+def has_session(view):
```
mircealungu, line 13: i see the wrapper as an annotation for a method. 

take for example @classmethod. It is something that we want to say *about the method itself*. 

Thus, ours could say something like:  
E.g. 

```python
@requires_user_session
def modify_user_language():
```
or `requires_logged_in` maybe? or simply `requires_session`? or `expects_session`?  

does this make sense to you?

zeeguu/api/utils/route_wrappers.py
```diff
@@ -19,16 +24,37 @@ def with_session(view):
     def wrapped_view(*args, **kwargs):
         try:
             session_uuid = flask.request.args["session"]
-            session_object = Session.find(session_uuid)
-            if session_object is None:
+            user_id, timeout = SESSION_CACHE.get(
+                session_uuid,
+                (
+                    None,
+                    datetime.now() - timedelta(0, SESSION_CACHE_EXPIRE),
```
mircealungu, line 31: is this not the default value? wouldn't it be more intutive to return None, None if we don't have anything in the cache? why do we compute some timeout in the past here? is it important to subtract exact the expiry from here? i might be tired :(

zeeguu/api/utils/route_wrappers.py
```diff
@@ -19,16 +24,37 @@ def with_session(view):
...
+                ),
+            )
+            if datetime.now() > timeout:
```
mircealungu, line 34: this is too complicated for me :(

by timeout I understand a duration, like 60s. so maybe this var should be named something like `session_expiry_time`?

zeeguu/api/endpoints/sessions.py
```diff
@@ -1,11 +1,30 @@
...
+from datetime import datetime
 from flask import request, make_response
 from zeeguu.core.model import Session, User
 from zeeguu.api.utils.abort_handling import make_error
 
-from zeeguu.api.utils.route_wrappers import cross_domain, with_session
+from zeeguu.api.utils.route_wrappers import cross_domain, has_session
 from . import api, db_session
 
+MAX_TIME_SESSION = 30  # Days
```
mircealungu, line 10: `DAYS_BEFORE_SESSION_EXPIRY` or similar maybe ?

zeeguu/api/endpoints/sessions.py
```diff
@@ -1,11 +1,30 @@
...
+        flask.abort(401)
+    is_session_too_old = (
+        datetime.now() - session_object.last_use
+    ).days > MAX_TIME_SESSION
+    if is_session_too_old:
+        print("Session was too old! - logging out!")
+        db_session.delete(session_object)
+        db_session.commit()
+        flask.abort(401)
+        return False
```
mircealungu, line 25: pycharm claims this code is unreachable. i guess abort is implemented as a throw?

zeeguu/api/endpoints/sessions.py
```diff
@@ -1,11 +1,30 @@
...
 from zeeguu.api.utils.abort_handling import make_error
 
-from zeeguu.api.utils.route_wrappers import cross_domain, with_session
+from zeeguu.api.utils.route_wrappers import cross_domain, has_session
 from . import api, db_session
 
+MAX_TIME_SESSION = 30  # Days
+
+
+def validate_session(session_object):
```
mircealungu, line 13: is what this method does better summarized by `update_session_if_not_expired`? 

although this is also not good enough: the method returns a True/False so this should be clear from it's name.  

I think this method does too many things, and we should split it maybe?

zeeguu/api/utils/route_wrappers.py
```diff
@@ -19,16 +24,37 @@ def with_session(view):
...
+                from zeeguu.api.endpoints.sessions import validate_session
+
+                print("----------- Updating Cache! -----------")
+                session_object = Session.find(session_uuid)
+                is_session_valid = validate_session(session_object)
```
mircealungu, line 39: I was under the impression that we were only extending the session lifetime in the `/validate` endpoint. 

if we do it here, we'll end up just having to worry more on the frontend since every endpoint might invalidate the session. 

let's simplify the code and our lives.

mircealungu (changes requested):
Great work Tiago. 
A few comments and questions and renames. 
Biggest ask is to try to move the session expiry only in /validate. Unless I'm missing something!

zeeguu/api/utils/route_wrappers.py
tfnribeiro: Yes, let's go with `requires_session` guess that's quite self descriptive.

zeeguu/api/utils/route_wrappers.py
tfnribeiro: It's a good point. The only reason I used a time delta was to not have to check for None in the conditional, so I just set to something that would trigger the update. I now get both None.

zeeguu/api/utils/route_wrappers.py
tfnribeiro: Indeed. Personally, I just see a timeout as when it should expire, but then `session_expiry_time `fits for both! Changed!

zeeguu/api/endpoints/sessions.py
tfnribeiro: So this method is responsible for validating the expire date based on last use of the session. So it checks against the date and if the session is too old it deletes the old session and returns false.

I guess I could not delete it and just return false and the user would have to re-log and create a new uuid, so that delete could be removed. It's just a way of cleaning up.

I have renamed the method `is_session_valid`.

zeeguu/api/utils/route_wrappers.py
tfnribeiro: And that is correct. We only extend the session expiry date in the `/validate` edpoint. What you see here is just extending the timeout. 

This only invalidates the session in the case that a user has kept a session in for an entire month without ever refreshing it, and at that point I do think we should "re-validate" the session. This was the reason why I moved the logic of `is_session_valid` to a method, but realistically in the frontend I think we should aim for having a timeout that means that in normal circumstances, we never "re-validate" sessions, and we only do it after the server resets or something similar. This is also to ensue that we could force users to re-login by clearing the session table.

zeeguu/api/utils/route_wrappers.py
```diff
@@ -19,16 +24,28 @@ def with_session(view):
...
+                user_id = session_object.user_id
+                SESSION_CACHE[session_uuid] = (
+                    user_id,
+                    datetime.now() + timedelta(0, SESSION_CACHE_TIMEOUT),
+                )
+            print("----------- Using Cache! -----------")
+            flask.g.user_id = user_id
+            flask.g.session_uuid = session_uuid
 
         except Exception as e:
```
mircealungu, line 50: I guess we should capture this one also with sentry?

zeeguu/api/utils/route_wrappers.py
```diff
@@ -19,16 +24,28 @@ def with_session(view):
...
-            zeeguu.core.model.db.session.add(session_object)
-            zeeguu.core.model.db.session.commit()
+            user_id, session_expiry_time = SESSION_CACHE.get(
+                session_uuid,
+                (
+                    None,
+                    None,
+                ),
+            )
+            if session_expiry_time is None or datetime.now() > session_expiry_time:
```
mircealungu, line 34: thanks for the refactoring. this is much more readable now!

zeeguu/api/endpoints/sessions.py
```diff
@@ -1,11 +1,29 @@
...
+    print("----------- Validating the Session based on Last use!! -----------")
+    if session_object is None:
+        return False
+    is_session_too_old = (
+        datetime.now() - session_object.last_use
+    ).days > DAYS_BEFORE_EXPIRE
+    if is_session_too_old:
+        print("Session was too old! - logging out!")
+        db_session.delete(session_object)
+        db_session.commit()
```
mircealungu, line 23: if the name of the method suggest a predicate, then we should not be doing user logout in this method. 

solution might be removing this method because it does too many things:
- checks for session object being None... that's an ocymoron
- checking for is_expired
- deleting session if too old

and introducing an `is_session_expired` method. the null check and the session deletion gcan be done more expilicitly elsewhere where they belong.  

do you see what I mean? if not, i can give a first stab to a refactoring in this direction!

mircealungu (approved):
two more minor asks @tfnribeiro

zeeguu/api/endpoints/sessions.py
tfnribeiro: Could you check now - I have tried to make it a bit more explicit.

mircealungu: TAK! It's merged and deployed. It reads quite nicely now :)

PR #155, opened 2024-05-28
FIX: Handle RSS feeds Publish Times

tfnribeiro:
I have noticed that the RSS feeds are not parsed as datetimes and for that reason they were always been set to be published at crawl time. I have now handled this, where the struct_time is converted to datetime before being normalized to server time.

### DIFF
--- a/zeeguu/core/feed_handler/feed_handler.py
+++ b/zeeguu/core/feed_handler/feed_handler.py
@@ -1,5 +1,7 @@
 from datetime import datetime, timezone, timedelta
 from zeeguu.core.util.time import normalize_to_server_time
+from datetime import datetime
+from time import mktime, struct_time
 
 
 class FeedHandler:
@@ -11,9 +13,17 @@ class FeedHandler:
         self.image_url_string = ""
 
     def get_server_time(self, article_date) -> datetime:
-        if type(article_date) is datetime:
-            return normalize_to_server_time(article_date)
-        return datetime.now()
+        try:
+            if type(article_date) is struct_time:
+                # Convert in case it is a struct_time obj
+                article_date = datetime.fromtimestamp(mktime(article_date))
+            if type(article_date) is datetime:
+                return normalize_to_server_time(article_date)
+        except Exception as e:
+            print(
+                f"Failed parsing into Datetime, using current date: '{article_date}', '{e}'"
+            )
+            return datetime.now()
 
     def extract_feed_metadata() -> None:
         """
--- a/zeeguu/core/feed_handler/feed_handler.py
+++ b/zeeguu/core/feed_handler/feed_handler.py
@@ -22,6 +22,7 @@ class FeedHandler:
             print(
                 f"Failed parsing into Datetime, using current date. Date was: '{article_date}', Failed with: '{e}'"
             )
+        finally:
             return datetime.now()
 
     def extract_feed_metadata() -> None:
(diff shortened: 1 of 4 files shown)


tfnribeiro:
I have checked, we do not crawl in the future! 

`article_downloader.py` checks if the date is in the future, and if so ignores that item from the feed.

However, we do see some RSS feeds where date is not correctly extracted, but the newspaper library does manage to get the text. Since we have the object anyway, we could check if the date parsed from the newspaper library is lower and if so, use that instead? 


Something like this:
image

```
    ## Compare the newspaper parsed time with the RSS feed time
    ## If this is older, then it's likely the correct date.
    normalized_np_publish_date = normalize_to_server_time(np_article.publish_date)
    oldest_pub_date = (
        published_datetime
        if published_datetime < normalized_np_publish_date
        else normalized_np_publish_date
    )

    # Create new article and save it to DB
    new_article = zeeguu.core.model.Article(
        Url.find_or_create(session, url),
        title,
        ", ".join(np_article.authors),
        np_article.text,
        summary,
        oldest_pub_date,
        feed,
        feed.language,
        htmlContent=np_article.htmlContent,
    )
```

zeeguu/core/feed_handler/feed_handler.py
```diff
@@ -11,9 +12,18 @@ def __init__(self, url: str, feed_type: int):
         self.image_url_string = ""
 
     def get_server_time(self, article_date) -> datetime:
```
mircealungu, line 14: could this be all moved to normalize_to_server_time? the only thing that it does besides that is 
- converts a potential struct_time to datetime (could belong to the normalize)
- returns a default value if something goes wrong in the conversion or normalization I guess? 

and then 'normalize' to server time, could it be called 'convert_to_server_time'? I'm a bit unsure of the 'normalize' appropriateness in this context.

mircealungu (changes requested):
minor naming question + possible removal of method.

zeeguu/core/feed_handler/feed_handler.py
tfnribeiro: Sure, I moved it inside the now called `convert_to_server_time`. What about the other comment of RSS feeds not getting the date? Should we then attempt to use the newspaper?

mircealungu: why not! the only thing that I don't understand is why "is lower" means is better :) oh, i see: this is to solve the situation of dates in the future? if we test the earlier date, then if one is in the future, and the other is correct, we'll be using the correct one? that's fine! maybe add a little comment.

tfnribeiro: Okay, that's a fair question but here is my thought process:

- At this point, it's not an article in the future - they are filtered out before.
- If the date is older, then it's probably because the RSS didn't carry any date, and was set datetime.now()
- So if the newspaper gets an older date, it means that it successfully parsed a date that wasn't today.  If it failed, then the date will be more recent than the parsed on and we keep the oldest (when we crawled the article in the first place).

I do get that we could have a situation where the RSS feed gives us a more recent date than the newspaper and then we would overwrite it, essentially it's almost an Heuristic -> **If we find an older date when parsing, we use the oldest date.**, as long as we are consistent, I think it would be OK.

mircealungu: That sounds fair enough. Add this explanation as comment in the code, and it should be good! Then ping me when I should merge :)

mircealungu:
Cool. Thanks for double checking!

PR #160, opened 2024-06-10
fix: re-enable archlens render

mathiasbk1:


mircealungu (approved):
Tak @mathiasbk1 ! 
So was the problem the fact that I used the wrong @schema?

mathiasbk1:
> Tak @mathiasbk1 ! So was the problem the fact that I used the wrong @Schema?
...Seems like it? 
I won't say I expected it, but it was the first thing I saw that looked off, and the pull request was to re-enable the diff action in order to do some further testing 👍 Glad that it works now!

mircealungu:
I'm a bit upset with myself for not having tried this earlier! There would have been so many months of evaluating it and getting new ideas for the future! Tak again!

PR #161, opened 2024-06-10
moved testing outside of the core

mircealungu:
what if we moved the tests outside of the core folder?


mircealungu:
@mathiasbk1 thanks 👍 🎉

Issue #157, opened 2024-06-07
remove from scheduler when calling /dont_use_in_exercises


mircealungu:
otherwise it's going to still appear even if it's not fit anymore

PR #162, opened 2024-06-11
Fixed FitForStudy flag

tfnribeiro:
Includes fixes for the issues:

## https://github.com/zeeguu/api/issues/159

`ExerciseOutcome.free_text_feedback` was outdated, and most of the logic was handled in the Update part of the BasicScheduler. Which means even if the flag is set to 0, the user still sees it in the exercises. 

Since this is an outdated method, we don't really use it anymore, but for completion I provided a solution that checks for the set of characters in the outcome. If it contains the "expected" outcome letters, then it's not considered other feedback.  

## https://github.com/zeeguu/api/issues/157

Whenever the user toggles the words that are scheduled to not be studied anymore they need to be deleted from the scheduler. This was not done with the new UserPreference calls and was added. This is done by checking if the bookmark has a schedule associated in `BasicSRSchedule` and deleting it from that table.

For consistency, I also set the user preference if they ever give feedback on a word - as it is synonymous to explicitly saying they don't want that word to be used anymore. 

## Other Changes:

I have also moved the correctness test to `ExerciseOutcome.is_correct` as that seems the place where we test for the outcome of an exercise, rather than a local test at the `BasicSRSchedule.update` method.

### DIFF
--- a/zeeguu/core/definition_of_learned/is_learned.py
+++ b/zeeguu/core/definition_of_learned/is_learned.py
@@ -3,6 +3,9 @@ CORRECTS_IN_DISTINCT_DAYS_FOR_LEARNED = 4
 
 def is_learned_based_on_exercise_outcomes(exercise_log):
     """
+    10-06-2024 -> OUTDATED, now words are considered learned after the user
+    has completed a learning cycle and depeding on how the productive flag is set.
+
     :return:
     """
     if exercise_log.is_empty():
--- a/zeeguu/core/definition_of_learned/is_learned.py
+++ b/zeeguu/core/definition_of_learned/is_learned.py
@@ -1,18 +1,34 @@
-CORRECTS_IN_DISTINCT_DAYS_FOR_LEARNED = 4
+from zeeguu.core.word_scheduling.basicSR.basicSR import NEXT_COOLING_INTERVAL_ON_SUCCESS
+
+LEARNING_CYCLE_LENGTH = len(NEXT_COOLING_INTERVAL_ON_SUCCESS)
 
 
-def is_learned_based_on_exercise_outcomes(exercise_log):
+def is_learned_based_on_exercise_outcomes(exercise_log, is_productive=True):
     """
-    10-06-2024 -> OUTDATED, now words are considered learned after the user
-    has completed a learning cycle and depeding on how the productive flag is set.
+    Checks if the user has reported the exercise as too easy or looks into the
+    streaks of this exercise log. Currently (14/06/2024), Zeeguu uses 2 cycles of
+    4 different days in spaced repetition to concider a word learned.
 
-    :return:
+    If a user has 2 streaks of 4, it means they have completed a 2 full cycles,
+    and therefor have learned the word.
+
+    The user also has the option of saying they do not want productive exercises.
+    In this case, we only need to have a single streak of 4 to consider the bookmark
+    learned.
+
+    :return:Boolean, the bookmark is learned based on the exercises or not.
     """
     if exercise_log.is_empty():
         return False
 
-    return (
-        exercise_log.last_exercise().is_too_easy()
-        or len(exercise_log.most_recent_correct_dates())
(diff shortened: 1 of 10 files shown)


zeeguu/core/definition_of_learned/is_learned.py
```diff
@@ -3,6 +3,9 @@
 
 def is_learned_based_on_exercise_outcomes(exercise_log):
     """
+    10-06-2024 -> OUTDATED, now words are considered learned after the user
+    has completed a learning cycle and depeding on how the productive flag is set.
+
```
mircealungu, line 8: should we remove this then? it seems it's only called from one other function `update_learned_status` which is only called from one test. We can remove this, the function and the test, and nothing's lost. Or?

zeeguu/core/model/exercise_outcome.py
```diff
@@ -28,6 +28,32 @@ class ExerciseOutcome(db.Model):
...
+        """
+        The user might have multiple attempts:
+            e.g. "WWWHHC",
+                 "WWTTHC"...
+        By checking if the characters in the answer are part
+        of the set of possible reports of the exercise, we can
+        conclude it is a user attempt.
+        """
+        allowed_attempts = set(["C", "W", "T", "S", "H"])
+        return allowed_attempts.issuperset(set([c for c in outcome]))
```
mircealungu, line 42: Beautiful! I didn't know that `issuperset` was a thing :)

zeeguu/core/model/exercise_outcome.py
```diff
@@ -28,6 +28,32 @@ class ExerciseOutcome(db.Model):
...
+
+    @classmethod
+    def is_correct(cls, outcome: str):
```
mircealungu, line 45: Yes! Thanks for moving this here. It belongs much better!

mircealungu (changes requested):
only one question: should we remove the deprecate method then?

mircealungu:
test

zeeguu/core/definition_of_learned/is_learned.py
tfnribeiro: I do think it's safe to remove, but I don't feel comfortable removing methods from the codebase without your approval!

tfnribeiro:
I think for this one, I can see that we have the long test method that tests the bookmark being learned by looking at the exercise log. We don't really do this anymore, but we could technically do it by detecting 2 4 streaks in a row, 1 for the Productive (if the setting is enabled) and the Receptive. Should I look into implementing this or do I just remove the method?

mircealungu:
> I think for this one, I can see that we have the long test method that tests the bookmark being learned by looking at the exercise log. We don't really do this anymore, but we could technically do it by detecting 2 4 streaks in a row, 1 for the Productive (if the setting is enabled) and the Receptive. Should I look into implementing this or do I just remove the method?

It would be nice to have a test for this too indeed :)

tfnribeiro:
I have created a `count_number_of_streaks`, which counts exercises that are correct in a row until either seeing an incorrect exercise or reaching to the number of intervals we have, which resets the count. 

We expect that a learned word has 2, 4 consecutives streaks (Productive & Receptive) and 1 if just Receptive to be considered learned. We cannot have any other occurrence at the current implementation as that would break the scheduler. 

If at some point we increase or decrease the cycle by altering the dictionary with the intervals this method would update as well. 

I have also included the method `is_learned_based_on_exercise_outcomes` as a method that can be called from a bookmark, which makes a bit more sense to me (as it's the bookmark which has all the relevant information, not an `ExerciseLog`).

NOTE: This implementation does not care about the Intervals between the correct answers - @mircealungu  should we do this? 

I tested with some of my bookmarks and this might also be a good way to evaluate how difficult a bookmark is for a user:

image

We can see that the first bookmark I had it wrong at least 17 times without having any streak, 4 times after 1 attempt correct. While in the 3 one we can see that after getting it wrong twice at the beginning of the cycles and then I completed the cycles.

PR #163, opened 2024-06-19
Fixed adding interests

tfnribeiro:
When updating the Session handling, the _SearchSubscription_ was not taking the db_session as the first paramenter.


mircealungu (approved):
tak for fix @tfnribeiro !

PR #170, opened 2024-07-26
Subscribed email searches

KlaraK94:
Added the endpoints for subscribing to email notifications for subscribed searches. 
Added two migration files, one for adding columns for the receive_email in search_subscription and Search. And one for deleting all entries in the tables search and search_subscription.


zeeguu/api/endpoints/search.py
```diff
@@ -212,3 +215,38 @@ def search_for_search_terms(search_terms, page: int = 0):
...
+def subscribe_to_email_search(search_terms):
+    """
+    A user can subscribe to email updates about a search
+    """
+
+    search = Search.find(search_terms)
+    user = User.find_by_id(flask.g.user_id)
+    receive_email = True
+    SearchSubscription.update_receive_email(db_session, user, search, receive_email)
+    Search.update_receive_email(db_session, search_terms, receive_email)
```
mircealungu, line 232: I do not understand this. Why does Search have a `receive_email` attribute? It should only be the Subscription that knows about this.

zeeguu/api/endpoints/search.py
```diff
@@ -212,3 +215,38 @@ def search_for_search_terms(search_terms, page: int = 0):
...
+    """
+    A user can unsubscribe to email updates about a search
+    """
+
+    search = Search.find(search_terms)
+    user = User.find_by_id(flask.g.user_id)
+    receive_email = False
+    SearchSubscription.update_receive_email(db_session, user, search, receive_email)
+    
+    Search.update_receive_email(db_session, search_terms, receive_email)
```
mircealungu, line 250: this should also not be needed!

zeeguu/core/emailer/subscribed_search_email.py
```diff
@@ -0,0 +1,47 @@
...
+    emailer.send()
+
+def send_subscription_emails():
+    all_subscriptions = SearchSubscription.query.filter_by(receive_email=True).all()
+    all_searches = Search.query.all()
+    current_datetime = datetime.now(datetime.timezone.utc)
+    previous_day_datetime = current_datetime - datetime.timedelta(days=1)
+
+    for subscription in all_subscriptions:
+        for search in all_searches:
```
mircealungu, line 31: you should be able to simply say `subscription.search` because the `search` is defined inside of the `SearchSubscription` class as a relationship!

zeeguu/api/endpoints/search.py
KlaraK94: I have added it to Search because you use search to show data in the UI. So if it is not added, I will not be able to see the changes made to the receive_email. Most endpoints use this, return json_result(search.as_dictionary()), they do not use search_subscription.

zeeguu/core/emailer/subscribed_search_email.py
KlaraK94: So like this instead?

for subscription in all_subscriptions:
        articles = article_search_for_user(subscription.user_id, 2, subscription.search, page = 0)
        new_articles_found = [article for article in articles if article.published > previous_day_datetime]

        if new_articles_found:
                user = User.find_by_id(subscription.user_id)
                send_mail_new_articles_search(user.email, search)

zeeguu/api/endpoints/search.py
mircealungu: Ah! In that case we should really refactor the UI to use Subscription instead of Search. I already observed that we should do that in the PR for the Web, but I thought we can delay it for after Wednesday. But in this case we should prioritize this straight away.

zeeguu/core/emailer/subscribed_search_email.py
mircealungu: Yes, use `subscription.search`. And also in your last line `send_mail_new_articles_search(user.email, search)` you'll probably have to use the same.

zeeguu/api/endpoints/search.py
mircealungu: Actually, we already have a meeting tomorrow @ 10am. We could use that too. But it might be a bit more than one hour.

KlaraK94: I would appreciate some help to fix this. Tomorrows meeting works fine for me, and I can do more than one hour.

mircealungu: let's do that then!

mircealungu (approved):
let's merge this!

PR #176, opened 2024-08-06
Combined keywords into a single email

tfnribeiro:
Combined the keywords into a single email body:

image

Should avoid spamming users emails if they have multiple searches.

### DIFF
--- a/tools/send_subscription_emails.py
+++ b/tools/send_subscription_emails.py
@@ -21,6 +21,7 @@ def send_mail_new_articles_search(to_email, search):
             " ",
             "You can find your subscriptions here: https://www.zeeguu.org/articles/mySearches"
             " ",
+            " ",
             "Cheers,",
             "The Zeeguu Team",
         ]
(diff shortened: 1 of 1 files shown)


mircealungu:
Should we also add a link to where the searches can be found? i.e. `https://www.zeeguu.org/articles/mySearches`?

mircealungu:
I see an A/B testing opportunity here: one approach just mentions that there are new articles. The other one provides titles and links too :)

tfnribeiro:
That's also a good idea! 

Added the link;

image

mircealungu (approved):
thanks @tfnribeiro . I've added one extra line-break before the "Cheers". We should also consider replacing `"Cheers"` with `cheers(flask.g.user.learned_language)` at some point in the future :) 

But for now it's good... Let's wait for @KlaraK94's review and then we can deploy.

KlaraK94:
Looks good! Nice, that the user will not get spammed, and I think the issue with my version - was that it was not using subscription.search.keywords :-)

mircealungu (approved):
ok. merging then.

PR #177, opened 2024-08-07
Allow tracking of user preferences in reader

tfnribeiro:
Add the settings of the reader to the user_preferences table.


mircealungu (approved):
this is a cool solution. thanks!

PR #179, opened 2024-08-15
Fix Politikens Feed

tfnribeiro:
Sometimes politikens feed will encode danish characters with strange characters, it only happens in the title which gets displayed in the home page for the users.

Difficult to test, as it only happens occasionally.

Would this be fine? I made the change just specifically for the feed with issues so it doesn't affect anything else.

### DIFF
--- a/zeeguu/core/content_retriever/article_downloader.py
+++ b/zeeguu/core/content_retriever/article_downloader.py
@@ -158,6 +158,11 @@ def download_from_feed(
                 url,
                 crawl_report,
             )
+            # Politiken sometimes has titles that have
+            # strange characters instead of å æ ø
+            if feed.id == 136:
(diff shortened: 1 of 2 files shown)


zeeguu/core/content_retriever/article_downloader.py
```diff
@@ -158,6 +158,11 @@ def download_from_feed(
                 url,
                 crawl_report,
             )
+            # Politiken sometimes has titles that have
+            # strange characters instead of å æ ø
+            if feed.id == 136:
```
mircealungu, line 163: this is a bit fragile in the face of the possibility of deleting and recreating the Politiken feed, case in which we would not have it on ID 136. But that's not going to happen any time soon I guess?

PR #181, opened 2024-08-20
Check for quality before performing clean-up

tfnribeiro:
This is the lines I think were the culprit.

If you look at the old `download_and_parse_with_remove_sents` the text would be stripped of all the repeating patterns that were used to distinguish if they are "paywalled" or not. 

To fix this, I renamed the method to reflect that we download and parse with readability, and then we text for the quality of the text before removing any patterns. 

One thing that I am not sure, is that now that we are storing the broken articles, should we then set the text to be what it was before the clean-up (in case we want to retrain the model later) or should we leave the clean text?

### DIFF
--- a/zeeguu/core/model/article.py
+++ b/zeeguu/core/model/article.py
@@ -281,7 +281,7 @@ class Article(db.Model):
         from zeeguu.core.model.article_broken_code_map import ArticleBrokenMap
 
         article_broken_map = ArticleBrokenMap.find_or_create(session, self, broken_code)
-        self.broken = BROKEN_ARTICLE_QUALITY
+        self.broken = MARKED_BROKEN_DUE_TO_LOW_QUALITY
         session.add(article_broken_map)
         session.add(self)
         session.commit()
@@ -301,7 +301,7 @@ class Article(db.Model):
         session.add(ua)
 
     def mark_as_low_quality_and_remove_from_index(self):
-        self.broken = BROKEN_ARTICLE_QUALITY
+        self.broken = MARKED_BROKEN_DUE_TO_LOW_QUALITY
         # if it was in ES, we delete it
         from zeeguu.core.elastic.indexing import remove_from_index
 
--- a/zeeguu/core/model/article.py
+++ b/zeeguu/core/model/article.py
@@ -22,6 +22,7 @@ article_topic_map = Table(
 )
 
 MAX_CHAR_COUNT_IN_SUMMARY = 300
+MARKED_BROKEN_DUE_TO_LOW_QUALITY = 100
 
 HTML_TAG_CLEANR = re.compile("<[^>]*>")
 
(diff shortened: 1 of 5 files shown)


tfnribeiro:
Merged the fixes to the bookmark change to pass unit testing in this change.

tfnribeiro:
@mircealungu  It still seems the render fails?

mircealungu:
yes, the render fails because of the version of `archlens.json` that's on that branch. 
I've only fixed it on `master`. 
it will stop failing from the next PR on.

tfnribeiro:
> yes, the render fails because of the version of `archlens.json` that's on that branch. I've only fixed it on `master`. it will stop failing from the next PR on.

I did merge master into this though - I thought that would include the fix?

mircealungu:
I'm confused :) 
Did you merge `tfnribeiro/master`? 
because there, indeed, the `"name"` in the schema still had a space in it. 
which was the problem.
but this is fixed now in the `zeeguu/api/master`

tfnribeiro:
> I'm confused :) Did you merge `tfnribeiro/master`? because there, indeed, the `"name"` in the schema of the `archlens.json` still had a space in it. which was the problem. but this is fixed now in the `zeeguu/api/master`

No, I did merge `zeeguu/api/master` into this branch - you can see the file has no changes when compared to the current state of `zeeguu/api/master`.

image

So, if it's fixed, I don't see why the pull wouldn't render the diagram.

tfnribeiro:
The reported error also seems related to URL access:`fatal: unable to access 'https://github.com/zeeguu/api/': The requested URL returned error: 403` not sure why it can't access the github link - maybe it's because it comes from a fork rather than the main branch?

tfnribeiro:
> Ah. That’s a different error than I’ve got. It might be related indeed to running from a fork. We must test this, and if it’s so, find a solution for it!
> […](#)
> On Aug 21, 2024 at 11:57 +0200, Tiago Ribeiro ***@***.***>, wrote: The reported error also seems related to URL access:fatal: unable to access 'https://github.com/zeeguu/api/': The requested URL returned error: 403 not sure why it can't access the github link - maybe it's because it comes from a fork rather than the main branch? — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were mentioned.Message ID: ***@***.***>

Seems like it's not possible: https://community.render.com/t/pull-request-previews-from-branch-on-fork/1953/3 so I don't know, it will always be a problem unless someone is working directly on this branch.

zeeguu/core/model/article.py
```diff
@@ -22,6 +22,7 @@
 )
 
 MAX_CHAR_COUNT_IN_SUMMARY = 300
+MARKED_BROKEN_DUE_TO_LOW_QUALITY = 100
```
mircealungu, line 25: sorry Tiago, but we don't seem to be using this constant anywhere. Or am I missing something?

mircealungu (approved):
looks good.

PR #182, opened 2024-08-20
Fix failing tests after change in Bookmarks

tfnribeiro:
When adding the consecutive_answers to the bookmarks we can run into exceptions if the bookmark as not been scheduled yet. In these cases, the value should be set to `Null` and this should fix the tests failing in the API. 

- We are not guaranteed that a bookmark has been trained and has consecutive answers. In this case, we should return null.


tfnribeiro:
Not sure why tests are failing? 

image

mircealungu (approved):

PR #184, opened 2024-08-21
Update archlens.json

mircealungu:


tfnribeiro:
To be fair, how are we supposed to read the diagram? Since you only changed one line in the json? What do the arrows mean?

mircealungu:
LOL. Good question. The actual diff is green for additions and red for deletions. There's none in this case... but the text should be clearer...

mircealungu:
and it should not even be rendered for the situation where there's no changes to this particular view... but I need to find somebody else to adopt this project and push it forward for this to be a thing!

PR #190, opened 2024-08-26
fix of subscribe to search - an open issue in the web #459

KlaraK94:
With the last changes, we made sure no duplications were added to search (after discussion with Mircea). But I do think duplications is necessary. The users can both add the same search subscription, but the issue is when removing the subscription again. Because there should then be two objects, where each user can delete "their own". But not having duplications results in only one object in the database, and neither/only one can remove the search.

This is a fix of an open issue in the web, https://github.com/zeeguu/web/issues/459.

### DIFF
--- a/zeeguu/core/model/search.py
+++ b/zeeguu/core/model/search.py
@@ -43,13 +43,10 @@ class Search(db.Model):
 
     @classmethod
     def find_or_create(cls, session, keywords):
-        try:
-            return cls.query.filter(cls.keywords == keywords).one()
-        except sqlalchemy.orm.exc.NoResultFound:
-            new = cls(keywords)
-            session.add(new)
-            session.commit()
-            return new
+        new = cls(keywords)
+        session.add(new)
+        session.commit()
+        return new
 
     @classmethod
     def find(cls, keywords):


tfnribeiro (commented):
With the last changes, we made sure no duplications were added to search (after discussion with Mircea). But I do think duplications is necessary. The users can both add the same search subscription, but the issue is when removing the subscription again. Because there should then be two objects, where each user can delete "their own". But not having duplications results in only one object in the database, and neither/only one can remove the search.

This is a fix of an open issue in the web, https://github.com/zeeguu/web/issues/459.

Thank you for investigating this Klara. Having duplication in a database is quite a bad idea, and after having a look myself you can see that we have a "mapping" table already, this is where we store the information of which users are interested in what searches (_search_subscription_). So the issue currently is the following:

```def unfilter_search():
    """
    A user can unsubscribe from the search with a given ID
    :return: OK / ERROR
    """

    search_id = int(request.form.get("search_id", ""))
    user = User.find_by_id(flask.g.user_id)
    try:
        to_delete = SearchFilter.with_search_id(search_id, user)
        db_session.delete(to_delete)
        to_delete = Search.find_by_id(search_id)
        db_session.delete(to_delete)
        db_session.commit()

    except Exception as e:
        log(str(e))
        from sentry_sdk import capture_exception

        capture_exception(e)
        return "OOPS. SEARCH AIN'T THERE IT SEEMS (" + str(e) + ")"

    return "OK"
```

If you look at the method above, we attempt to delete the Search once we delete the Search as soon as we delete the subscription. However, without duplication this will fail, because there can be other users subscribed to the same topic ('OL') in this case.

In order to delete the search, we need to ensure there are no users with that keyword subscribed to. I have made a change where this is done and I believe that's a better way to fix it. 

Take a look at https://github.com/zeeguu/api/pull/191 which fixes this issue - tested with 2 users in my local environment.

KlaraK94:
It makes sense to not have duplications in the code. Thank you for the fix!

PR #191, opened 2024-08-26
Fix users not able to delete search

tfnribeiro:
- We were trying to delete a search without checking if there were no users subscribed to it - Failing due to FK constraints.


mircealungu (approved):
tak @tfnribeiro !

PR #206, opened 2024-09-11
198 students invisible when joining second classroom

tfnribeiro:
Change to support the UI changes to allow multiple classrooms for one user (https://github.com/zeeguu/web/pull/526)

## Changelist

- Removed the `cohort_id` from the user table, instead a mapping table was introduced to track this relationship.
- The script first inserts all the `cohort_id` into the mapping table, so users remain in their current classrooms.
- Testing membership of users is now done by a method `is_member_of_cohort` which expects an ID and checks that mapping exists for the user.
- `basic_user_info` now sends a list of cohorts if the user is in multiple classrooms.
- Change methods to call their new equivalent, overall maintaining the same logic the end-points used to have. This mostly means that now we need to check for multiple entries and find if a specific ID is present.
- TeacherDashboard now reflects the time spent on exercises for the specific language, rather than all exercises done by a student. 

### Question:
(Before commiting @mircealungu)
- When we add/remove a user from a cohort, should we commit immediately or add it to the session and then the developer can decide when to commit the change?

### DIFF
--- a/zeeguu/api/endpoints/teacher_dashboard/student.py
+++ b/zeeguu/api/endpoints/teacher_dashboard/student.py
@@ -26,13 +26,14 @@ from zeeguu.core.model import db
 @api.route("/basic_user_info/<id>", methods=["GET"])
 @requires_session
 def basic_user_info(id):
+    # This method doesn't seem to be used by the frontend
     user = check_permission_for_user(id)
 
     return jsonify(
         {
             "name": user.name,
             "email": user.email,
-            "cohort_ids": [uc.id for uc in user.cohorts],
+            "cohorts": [uc.cohort.get_cohort_info() for uc in user.cohorts],
         }
     )
 
--- a/zeeguu/api/endpoints/teacher_dashboard/student.py
+++ b/zeeguu/api/endpoints/teacher_dashboard/student.py
@@ -29,7 +29,11 @@ def basic_user_info(id):
     user = check_permission_for_user(id)
 
     return jsonify(
-        {"name": user.name, "email": user.email, "cohort_id": user.cohort_id}
+        {
+            "name": user.name,
+            "email": user.email,
+            "cohort_ids": [uc.id for uc in user.cohorts],
+        }
     )
 
 
@@ -37,7 +41,7 @@ def basic_user_info(id):
 @requires_session
 def user_info_api():
     user, cohort, from_date, to_date = _get_student_cohort_and_period_from_POST_params()
-    check_permission_for_user(user.id)
+    check_permission_for_user(user.id, cohort.id)
 
     return jsonify(student_info_for_teacher_dashboard(user, cohort, from_date, to_date))
 
--- a/zeeguu/api/endpoints/user.py
+++ b/zeeguu/api/endpoints/user.py
@@ -8,7 +8,7 @@ from zeeguu.core.model import User
 from zeeguu.api.utils.json_result import json_result
 from zeeguu.api.utils.route_wrappers import cross_domain, requires_session
 from . import api
-from ...core.model import UserPreference, UserActivityData, UserArticle, Article
+from ...core.model import UserActivityData, UserArticle, Article
 
 
 @api.route("/learned_language", methods=["GET"])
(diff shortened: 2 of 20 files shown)


mircealungu:
> When we add/remove a user from a cohort, should we commit immediately or add it to the session and then the developer can decide when to commit the change?

are you talking about this? 
https://github.com/zeeguu/api/pull/206/files#diff-bb192734c50e63f5c18ff35dcab929883ceadefef56fa5c32fef107d21bf7d2eR321

It s ok how you're doing it. Although every commit takes a bit of time.

We could try to create a rule of thumb that says that something like: 

> things should be committed by the caller if possible

or

> if a method gets a session, it should modify it, but not commit by default

So in this case then the `add_user_to_cohort` would just add the user to the session but leave the commit to be done by the parent.

However, I'm not sure that there won't be exceptions from these rules.

zeeguu/api/endpoints/user.py
```diff
@@ -198,3 +198,24 @@ def send_feedback():
...
+
+
+@api.route("/leave_cohort/<cohort_id>", methods=["GET"])
+@cross_domain
+@requires_session
+def leave_cohort(cohort_id):
+    from zeeguu.core.model import db
+
+    """
+    set the native language of the user in session
```
mircealungu, line 210: this comment is off?

zeeguu/api/endpoints/teacher_dashboard/student.py
```diff
@@ -26,18 +26,23 @@
 @api.route("/basic_user_info/<id>", methods=["GET"])
 @requires_session
 def basic_user_info(id):
+    # This method doesn't seem to be used by the frontend
```
mircealungu, line 29: really? then maybe we should remove it?

zeeguu/core/model/user.py
```diff
@@ -123,6 +117,18 @@ def create_anonymous(
...
+        cohort_id = int(cohort_id)
+        return any([c.cohort_id == cohort_id for c in self.cohorts])
+
+    def remove_from_cohort(self, cohort_id):
+        from zeeguu.core.model.user_cohort_map import UserCohortMap
+
+        cohort_id = int(cohort_id)
+        UserCohortMap.query.filter(UserCohortMap.user_id == self.id).filter(
+            UserCohortMap.cohort_id == cohort_id
+        ).delete()
```
mircealungu, line 130: Whoa! you can do that w/o a session :)

mircealungu (approved):
that was quite a bit of work.
looks good overall.
minor asks and we're ready to deploy.

zeeguu/api/endpoints/teacher_dashboard/student.py
tfnribeiro: image

Seems that way, I believe it was something that we didn't end up using when Iga did the onboarding.

zeeguu/api/endpoints/user.py
tfnribeiro: I didn't edit it, seems like it has been this way for 6 years - but I can remove it - I think it came in the diff due to the formatting.

mircealungu:
> > > When we add/remove a user from a cohort, should we commit immediately or add it to the session and then the developer can decide when to commit the change?
> > 
> > 
> > are you talking about this? https://github.com/zeeguu/api/pull/206/files#diff-bb192734c50e63f5c18ff35dcab929883ceadefef56fa5c32fef107d21bf7d2eR321
> > It s ok how you're doing it. Although every commit takes a bit of time.
> > We could try to create a rule of thumb that says that something like:
> > > things should be committed by the caller if possible
> > 
> > 
> > or
> > > if a method gets a session, it should modify it, but not commit by default
> > 
> > 
> > So in this case then the `add_user_to_cohort` would just add the user to the session but leave the commit to be done by the parent.
> > However, I'm not sure that there won't be exceptions from these rules.
> 
> Yes, it's just that in the code you can see in the addition I do the commit in the method in add user, while in the delete I am doing it in the API call rather than the User.Method, where it just adds the value to be deleted to the session, it can then be deleted at the API level. I think that it creates a bit of a conflict in the semantics of add/delete user, if the user is not really added or deleted. What do you think we commit, and if we decide to use sessions, maybe we should call the method "mark_for_delete/add" - so it's more explicit it doesn't delete the user right away.

this is an interesting idea, but i wonder if it won't make our method names too long? 

```
    def mark_for_add_user_to_cohort(self, cohort, session):
        from zeeguu.core.model.user_cohort_map import UserCohortMap

        new_cohort = UserCohortMap(user=self, cohort=cohort)
        session.add(new_cohort)
```

the more we discuss the more I think that somebody who gets a session should not commit it. 
you get a session you add to it, and that's it. 

```
    def add_user_to_cohort(self, cohort, session):
        from zeeguu.core.model.user_cohort_map import UserCohortMap

        new_cohort = UserCohortMap(user=self, cohort=cohort)
        session.add(new_cohort)


alternative would be to not pass the session at all anymore, but instead such methods that aim to modify the DB model always return the objects that have to be added to the session and committed? 

e.g. 

```
    def add_user_to_cohort(self, cohort, session):
        from zeeguu.core.model.user_cohort_map import UserCohortMap

        new_user_cohort_map = UserCohortMap(user=self, cohort=cohort)
        return new_user_cohort_map
```

do you see what I mean?

tfnribeiro:
I do see it - and generally I feel like it makes more sense that things are added rather than commited immediately especially if we were to do bulk updates, then it would be much better to store it in session. I think returning the altered object could be a good solution for this - and we can slowly start implementing it.

Should I create an issue where we can track it?

mircealungu:
Just to be on the same page, I still think the best approach would be to simply pass the session, and add things to it. To be sure that we're on the same page with this, I've created a Discussion category called [Coding Guidelines](https://github.com/zeeguu/api/discussions/categories/coding-guidelines) and also a separate file [Coding Conventions](https://github.com/zeeguu/api/blob/master/CODING-GUIDELINES.md).

PR #207, opened 2024-09-12
205 frontpage remains static for long

tfnribeiro:
## Changes

Re-worked the priority given to the different parameters in the query.
- In the home page, we no longer use the should query to include the user searches, but instead do a small query to find if there are any relevant new articles which are then ordered with the general topic preferences and recency filter. The home page essentially prioritizes Topic > (Recency == Difficulty Level)
- Extended the Search endpoint to allow users to fine-control their searches. The user can use the UI to specify they'd rather see more relevant content at the cost of less recent articles or difficulty. 

The UI changes are implemented in:  https://github.com/zeeguu/web/pull/530

### DIFF
--- a/zeeguu/api/endpoints/search.py
+++ b/zeeguu/api/endpoints/search.py
@@ -268,7 +268,7 @@ def search_for_latest_search_terms(search_terms):
     user = User.find_by_id(flask.g.user_id)
     articles = article_search_for_user(
         user,
-        5,
+        3,
         search_terms,
         page=0,
         use_published_priority=True,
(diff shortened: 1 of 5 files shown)


zeeguu/api/endpoints/search.py
```diff
@@ -216,8 +221,60 @@ def search_for_search_terms(search_terms, page: int = 0):
...
+# ---------------------------------------------------------------------------
+@api.route(f"/{LATEST_SEARCH}/<search_terms>", methods=("GET",))
+# ---------------------------------------------------------------------------
+@cross_domain
+@requires_session
+def search_for_latest_search_terms(search_terms):
+    """
+    This endpoint is used for the preview search and sending notification
+    emails.
+    The call prioritizes the recency of articles paried with the difficulty.
```
mircealungu, line 258: what is "paried"?

PR #211, opened 2024-09-13
204 gmail detects the text contents to be the same and hides them in continuous emails

tfnribeiro:
Updated the text sent in the Notification Emailer so the user can see the titles of the new articles that have arrived. The articles themselves have an hyperlink to the website where they were found so the user can go there immediately. 

@mircealungu Would it be possible to just test it to see if the HTML works as expected before fully deploying it?


mircealungu (approved):
Cool!

PR #213, opened 2024-09-13
196 report should we have subscribed topics report

tfnribeiro:
Added tracking to what Keywords are being filtered / added in order to add sources that fit our users preferences.

image


mircealungu (approved):
nice. the more data in the report, the better :)

PR #217, opened 2024-09-17
Lower the threshold to show read articles.

tfnribeiro:
- I realized the units weren't correct in the check. It should be in decimals (percentage) rather than int. - I can still see that sometimes I finish an article but it's still popping up in my feed, so maybe lowering it to 0.9 might be a good idea. @mircealungu has this ever occurred to you?


mircealungu:
I didn't observe this happening. But I'll merge in the hope of not observing it in the future :)

PR #236, opened 2024-10-02
Support Mock Translations for Development

tfnribeiro:
I have found myself using this quite often in development to have exercises with new test users, without having to rely on existing users in the frontend. It's also useful to test the styling for translations among other functionality.

## Changes:

- Changed the docker-compose to include an extra environment variable: `DEV_SKIP_TRANSLATION`
- `get_one_translation(from_lang_code, to_lang_code)` now will check if the flag `DEV_SKIP_TRANSLATION` is enabled, if so it will generate translations that follow the format T-([translation-language])-'[original_word]'. E.g. "hus" would return T-EN-'hus' if the translation language is English.

## Preview:

{EBE170A0-6221-42F5-BB27-7AF50EF4C728}

{4DD777B5-F6C4-421B-90B4-653B9F76A09C}


mircealungu:
brilliant solution with t-(en)-forskere :)

zeeguu/api/endpoints/translation.py
```diff
@@ -40,7 +41,7 @@ def get_one_translation(from_lang_code, to_lang_code):
 
     :return: json array with translations
     """
-
+    IS_DEV_SKIP_TRANSLATION = int(os.environ.get("DEV_SKIP_TRANSLATION", 0)) == 1
```
mircealungu, line 44: should we move this at the top of the file? otherwise we're calling it every time we call the method. and it should not be needed: the envvars are unlikely to change during the running of the process.

mircealungu: the fact that it's all uppercase makes it even more ready to be a constant at the top :)

mircealungu (approved):
looks super cool. 
small comment and we merge!

zeeguu/api/endpoints/translation.py
tfnribeiro: Indeed, I am not sure why it was in the method itself. I have moved it now!

mircealungu (approved):
tak!

PR #243, opened 2024-10-03
CEFR Level shouldn't be necessary in user settings

tfnribeiro:
- If no CEFR level is passed in the settings then the settings update to the user is not run. The reason for this was because we had issues before where if a user would switch languages the field for their level might go empty and that would cause some data corruption. Now, settings have been isolated and it's not possible to change language without setting a level. Meaning that this constraint shouldn't be in place (or at least should not require the frontend to always send the cefr_level).

### Changes:

- Instead of cancelling the user update if cefr_level isn't provided, instead no update is made to the learned_language if no cefr_level is provided.


mircealungu (approved):
cool!

PR #246, opened 2024-10-05
Add 0 default to max and min id

benceluzsinszky:
The script crashes if the DB is empty, as max_id and min_id is None and the loop can't iterate on them.


mircealungu (approved):
thanks for the fix... in reality if you don't have a DB then you should not try to run mysql_to_elastic in the first place :) but let's fold it in, it can't hurt.

PR #250, opened 2024-10-07
Script for recomputing FK scores for languages with new constants

gustavkrist:
I've added a script to recompute FK scores for all the languages that just had new constants added, similar to https://github.com/zeeguu/api/blob/master/tools/old/recompute_fk_difficulties_for_polish.py.


tfnribeiro:
I also think it's good!

I would just add a checkpoint so commit every 1000 articles or so - just so we don't have a really large transaction.

tfnribeiro:
I was attempting to run it in my environment and the process seems to get killed about a 1/3 into the process. 

{B5AEBBB1-C57B-4073-931C-1EC001988DCF}

Checkpointing in-between seems to allow the process to continue past that point:

{DC4FDF23-6EAF-4A4B-B4B0-196DC35A3513}

PR #261, opened 2024-10-21
Add Error Reporting for ES Inserts + Add URL Keywords to Report

tfnribeiro:
- Improved the error reporting in the bulk inserts by printing the errors found and the total Success + Errors
- Added the URL keywords to report where any URL keywords seen more than 100 times with no Topic assigned will be shown in the table. 

Errors might have been due to the attempt to index broken articles with empty fields, which fail when adding to ES. To avoid these situations, I also filter the broken articles in the query now. I expect less errors now, but if we do we should get them printed out with some information about what was the cause.

### DIFF
--- a/tools/mysql_to_elastic_new_topics.py
+++ b/tools/mysql_to_elastic_new_topics.py
@@ -26,8 +26,8 @@ app.app_context().push()
 DELETE_INDEX = False
 # First we should only index with topics so we can do
 # inference based on the articles that have topics.
-INDEX_WITH_TOPIC_ONLY = False
-TOTAL_ITEMS = 50000
+INDEX_WITH_TOPIC_ONLY = True
+TOTAL_ITEMS = 10000
 ITERATION_STEP = 1000
 
 print(ES_CONN_STRING)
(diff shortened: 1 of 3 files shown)


tools/mysql_to_elastic_new_topics.py
```diff
@@ -26,8 +26,8 @@
 DELETE_INDEX = False
 # First we should only index with topics so we can do
 # inference based on the articles that have topics.
-INDEX_WITH_TOPIC_ONLY = False
-TOTAL_ITEMS = 50000
+INDEX_WITH_TOPIC_ONLY = True
+TOTAL_ITEMS = 10000
```
mircealungu, line 30: why is total items 10K?

mircealungu (approved):

tools/mysql_to_elastic_new_topics.py
tfnribeiro: Because I didn't want to run it for all documents in my environment. I can add a comment that number should be adjusted as needed, I guess it could even be made into a system argument.

mircealungu:
Comment suffices!

PR #266, opened 2024-10-28
Remove bookmark.learned boolean as its redundant with learned time

tfnribeiro:
- Replaced all the instances where bookmark.learned was used. Now, we check for the datetime in learned_time, if it exists, then the bookmark has been learned.
- Tested the different endpoints (execises, history, updating the bookmark learned status) and it results in expected behaviour.


mircealungu (approved):
looks good. let's merge!

PR #283, opened 2024-11-05
281 vejret shouldnt be classify as technology science

tfnribeiro:
As described in the issue, when labeling I thought vejret would have news that are similar to what's associated with the keyword 'klima', which are more focused on larger climate events rather than description of the weather in a daily / weekly basis that we see in 'vejret'.

This means we should remove this to avoid these types of news to be inferred as 'Science & Technology' in the future.

I added a script (tools/update_es_based_on_url_keyword.py), where we can give a keyword to delete the current mappings topics with that keyword, and re-index the articles accordingly. This can also be used to simply update a keyword which had no mapping and has been mapped a topic.


tfnribeiro:
After a discussion with Mircea, we identified some edge-cases that needed to be covered correctly.

I have made a series of improvements, and I will exemplify an example here, currently I have "vejret" associated with the topic "news", and I want to remove this topic.

image

I will run the script with the following options:

```
URL_KEYWORD_TO_UPDATE = "vejret"
DELETE_ARTICLE_NEW_TOPICS = True
RECALCULATE_TOPICS = True
RE_INDEX_ONLY_ARTICLES_IN_ES = True
ITERATION_STEP = 1000
```

This means, I want to delete the mapping "vejret" -> "News", and recalculate the topics for those articles. 

This results in the following output:

```
Started at: 2024-11-08 13:06:32.807006
Got articles with url_keyword 'vejret', total: 650
Deleting new_topics 'News' for articles which have the keyword: 'vejret'
Found '650' topic mappings to delete.
MySQL deletion completed.
Re-indexing only existing articles in ES...
Starting re-indexing process...
  0%|                                                                                 | 0/1 [00:00<?, ?it/s]
Batch finished. ADDED/UPDATED:650 | ERRORS: 0
100%|█████████████████████████████████████████████████████████████████████████| 1/1 [00:20<00:00, 20.68s/it]
[]
Total articles added/updated: 650
Ended at: 2024-11-08 13:07:26.641374
Process took: 0:00:53.834368
```

Now, going back to the home page:

image

The articles are no longer stored as "News", but can still be found by searching:

image

@mircealungu Take a look when you have time and let me know if it makes more sense now!

tools/update_es_based_on_url_keyword.py
```diff
@@ -0,0 +1,172 @@
...
+                yield (article)
+            except NoResultFound:
+                print(f"fail for: '{i}'")
+            except Exception as e:
+                print(f"fail for: '{i}', {e}")
+
+    def gen_docs(articles_w_topics):
+        for article in articles_w_topics:
+            try:
+                yield create_or_update_bulk_docs(article, db_session)
```
mircealungu, line 94: why is this called `bulk_docs`? it seems to work on a single article.

tools/update_es_based_on_url_keyword.py
```diff
@@ -0,0 +1,172 @@
...
+
+def main():
+    def fetch_articles_by_id(id_list, recalculate_topic=False):
+        for i in id_list:
+            try:
+                article = Article.find_by_id(i)
+                if not article:
+                    print(f"Skipped for: '{i}', article not in DB.")
+                    continue
+                if recalculate_topic:
```
mircealungu, line 82: this is not clean code. a function should do one thing and one thing only. here the function is `fetch`, a getter, but then it also does some other commits to the DB. Could we separate these? I wonder if that would help me better understand the code.

tools/update_es_based_on_url_keyword.py
```diff
@@ -0,0 +1,172 @@
...
+            NewArticleTopicMap.article_id.in_(list(target_ids))
+        ).filter(NewArticleTopicMap.new_topic_id.in_(new_topics_ids_to_delete))
+        print(
+            f"Found '{len(topic_mappings_to_delete.all())}' topic mappings to delete."
+        )
+        topic_mappings_to_delete.delete()
+        db_session.commit()
+        print("MySQL deletion completed.")
+
+    if len(target_ids) == 0:
```
mircealungu, line 130: should this be moved above, immediately after the getting of the `target_id`s? the `print('Got articles...` is on the other branch of this if I think.

tools/update_es_based_on_url_keyword.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    print(
+        f"Got articles with url_keyword '{URL_KEYWORD_TO_UPDATE}', total: {len(target_ids)}",
+    )
+
+    # Updating url_keyword new_topic mapping
+    # And the topics that were added based on that keyword.
+    if DELETE_ARTICLE_NEW_TOPICS:
+        new_topics_ids_to_delete = []
```
mircealungu, line 108: is this precisely named? do we ever delete a `topic_id`?

mircealungu (changes requested):
a few questions for you @tfnribeiro

tools/update_es_based_on_url_keyword.py
tfnribeiro: No, we don't delete topic_ids, but we delete any articles that contain the topics_ids that were associated with the keyword. 

I tried to name it 'topics_ids_to_delete_mappings', hopefully that makes more sense.

tools/update_es_based_on_url_keyword.py
tfnribeiro: I don't think so. This has to do with re-indexing, not the data integrity. If I moved this above, then there would be data that wouldn't be updated in the database, which I think is more confusing. 

The way I see it, is that the DB should always be consistent regarding the mappings, but the ES can be incomplete, e.g. some articles might not be in ES and we do not necessarily need to index everything in ES.

tools/update_es_based_on_url_keyword.py
tfnribeiro: It is an iterator that works for each article and creates a json to work for the bulk function, so the bulk here means that it's to be used by the bulk functions, not necessary that is taking a bulk of articles. I renamed it to: 'create_or_update_doc_for_bulk' maybe that's a bit more clear?

tools/update_es_based_on_url_keyword.py
tfnribeiro: I moved the recalculate topics to the function: 'recalculate_article_url_keyword_topics', which is run before the indexing step.

tools/update_es_based_on_url_keyword.py
tfnribeiro: With some testing, I realized that originally, I was only updating the articles in the DB that were in ES. I still think we should keep the DB data correct with the updates, even if it takes somewhat longer - but ES can be a bit more flexible. This was related with the 'fetch_articles_by_id' not being clean, and re-calcualting the topics when fetching. 

I think the flow now makes more sense: Delete Mappings and the Associating of the Keyword -> Recalculate the Topics for Articles -> Index them in ES.

tfnribeiro:
I have reviewed the code, and I hope now it's more understandable. I have renamed some variables / methods, and tried to make the steps more clear.

Let me know if it is more clear?

tfnribeiro:
I also fixed a bug introduced by the quicker method of filtering articles in ES. This only happens when the user doesn't have the ES index zeeguu. Overall, it should be more robust this way.

mircealungu (approved):

PR #293, opened 2024-11-11
updated comment

mircealungu:
@tfnribeiro - i found this change uncommitted on my machine.
please double-check that the comment is correct. tka!


tfnribeiro (approved):
Yes, this is true - can be merged.

PR #297, opened 2024-11-14
Update generate_report.py

tfnribeiro:
- Added a way to visualize how many of the topics are inferred / assigned by url_keywords.


mircealungu (approved):
more report info, more better :)

PR #298, opened 2024-11-15
Removing old topics, renaming new topics

tfnribeiro:
- 'Removal' of old topics, and localized topics, with renaming the `new_topics` to `topics`
- Moved `es_v8_migration` folder to `old`, as we have made this migration, though the scripts might be worth keeping for a while longer.
- Updated the Report to not include old vs new topics
- Remove the check for version of ES7 vs ES8
- Updated tests to reflect the new topic inference

I am creating this as a draft as we need to discuss how to deploy it. The frontend also needs changes that are implemented in: https://github.com/zeeguu/web/tree/remove-old-new-topics-distinction, otherwise the frontend won't render the articles.


mircealungu:
ehi @tfnribeiro - this is massive work!

minor: there seems to be a conflict in `generate_report`? 

regarding deployment... we're small enough that we can afford having the app down for a few minutes? 
if you look at hourly utilization in the FMD dashboard you'll see that we can likely do a 7am migration and not many people would be affected. 

image

otherwise we'd have to add transient code in the frontend that works with both the old and the new api, migrate the api, then remove the transient code. i think it's too much and not needed at the moment. what do you think? see you tomorrow at 7 online? :)

tfnribeiro:
I will fix the conflict, I guess it has to do with one of the PRs we just merged. I can log in tomorrow at 7AM and we can do it then.

mircealungu (approved):

PR #301, opened 2024-11-19
New level scheduling

merleschoen:
With this pr we introduce a new scheduling algorithm with four difficulty levels. 

| Level  | Exercises |
| ------------- | ------------- |
| 1  | Match, Multiple-Choice L1, Click Word In Context  |
| 2 | Multiple-Choice L2, Multiple-Choice Context, Multiple-Choice Audio |
| 3 | Translate What You Hear, Translate Into L1  |
| 4 | Spell What You Hear, Translate Into L2  |

A bookmark is now considered learned once it has progressed through cooling interval 0 -> 1 -> 2 for all four levels. 
This pr also includes a refactoring of the BasicSR class into a parent class with two children: LevelsSR and LearningCycleSR, so it will be easier for us to introduce new schedulers in the future.

Currently, the levels are implemented as a feature that is only available to selected classes.


zeeguu/core/test/test_bookmark.py
```diff
@@ -18,6 +14,8 @@
 from zeeguu.core.test.rules.user_rule import UserRule
 from zeeguu.core.model import Bookmark
 from zeeguu.core.model import db
+from zeeguu.core.word_scheduling.basicSR.learning_cycle_SR import LearningCycleSR
+from zeeguu.core.word_scheduling.basicSR.basicSR import ONE_DAY
```
tfnribeiro, line 18: This is unused.

zeeguu/core/word_scheduling/basicSR/learning_cycle_SR.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        in_days:bool False, use true if you want the interval in days, rather than
+        minutes.
+        :returns:int, total number of minutes the schedule can have as a maximum.
+        """
+        return cls.MAX_INTERVAL if not in_days else cls.MAX_INTERVAL // ONE_DAY
+
+    @classmethod
+    def get_next_cooling_interval(cls):
+        return cls.NEXT_COOLING_INTERVAL_ON_SUCCESS
```
tfnribeiro, line 93: Same as in levels_SR

zeeguu/core/word_scheduling/basicSR/learning_cycle_SR.py
```diff
@@ -0,0 +1,146 @@
...
+            # Article.
+            return
+        schedule.update_schedule(db_session, correctness)
+        db_session.commit()
+
+    @classmethod
+    def find_or_create(cls, db_session, bookmark):
+
+        results = cls.query.filter_by(bookmark=bookmark).all()
+        print(results)
```
tfnribeiro, line 130: Do we need this print?

tfnribeiro (commented):
Looks good overall, I have some comments as there is one method that seems a bit confusing without a name change.

zeeguu/api/endpoints/feature_toggles.py
```diff
@@ -79,7 +79,7 @@ def _merle_exercises(user):
 def _exercise_levels(user):
     right_user = (
         user.invitation_code == "exercise_levels"
-        or user.is_member_of_cohort in [437, 514, 451, 519, 475, 506, 523]
+        or user.is_member_of_cohort in [514, 451, 523, 495, 489, 519, 511, 518]
```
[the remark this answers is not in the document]
merleschoen: Yes, we have 85 testers and 84 users in the control group. Unfortunately, it's not really possible to perfectly balance the language and experience with the app, but both groups contain users that have been using the app for over a year and users that have started using it this summer. I've also made sure to have both German and French users in both groups.

tfnribeiro:
I had a look at the logic here and there was only a small bug where the count to the consecutive_correct_answers wasn't getting updated when we upgraded level and the cooling interval update was getting overriden, also when upgrading level.

Other than that, I think it's OK! My only other suggestion was to handle the situation where we transform the LearningCycle -> Levels to also set the LearningCycle to 0, so we could do something similar if we went from Levels->LearningCycle, but that shouldn't happen too much, so we can just leave it as it is for now.

tfnribeiro:
My last suggestion would be to add a UnitTest to each of these schedules - I think at this point we are getting a bit of complicated logic that we should track if it breaks at some point.

tfnribeiro:
@mircealungu Seems like the azure dependency is breaking the tests, we might have to make a new image for docker with the dependency installed.

mircealungu:
@tfnribeiro - should be fixed now.

tfnribeiro:
@mircealungu I have written the tests and given a test to the frontend and things look OK! 

Since I had to change some of the functions to support the tests, I had to fix some bugs that resulted from that, but I think now everything should be good. When you have a chance take a look at the tests let me know what you think.

zeeguu/core/definition_of_learned/is_learned.py
```diff
@@ -21,14 +24,28 @@ def is_learned_based_on_exercise_outcomes(exercise_log, is_productive=True):
...
+
+    learning_cycle_length = len(scheduler.get_cooling_interval_dictionary())
+
+    streaks_by_length = exercise_log.exercise_streaks_of_given_length()
+    full_cycles_completed = streaks_by_length.get(learning_cycle_length)
+
     if is_productive:
         return full_cycles_completed == 2
     else:
         return full_cycles_completed == 1
```
mircealungu, line 43: i've been trying to understand this function @tfnribeiro, and in the process also renamed and reformatted a bit the `exercise_streaks_of_given_length`. However, I don't fully understand it. Are we sure it's not buggy? 

In particular this conditional here, don't we need a full_cycles_completed to be 4 for the `Levels` scheduler? Or do I mis-understand what full_cyucles_completed means?

mircealungu (approved):
titanic work in the tests @tfnribeiro :) 
i've made a few refactorings, and with the exception of one curious possible bug, everything looks good.

PR #309, opened 2024-11-29
more human readable updates for exercise sessions

mircealungu:
As far as I can tell, even if I define the configuration to be 

```
MULTI_LANG_TRANSLATOR_AB_TESTING=False
```

the current code makes it True anyway. 

the current PR should fix that.


tfnribeiro (approved):
Looks good!

PR #310, opened 2024-11-30
minitool to show feeds for a given language

mircealungu:


tools/feeds_for_language.py
```diff
@@ -0,0 +1,23 @@
+from zeeguu.api.app import create_app
+
+from zeeguu.core.model import Language, Feed
+
+app = create_app()
+app.app_context().push()
+
+language = Language.find(input("Language code: "))
+all_italian_feeds = Feed.query.filter_by(language_id=language.id).all()
```
tfnribeiro, line 9: Variable name should be generic, `all_feeds_for_language`?

tools/feeds_for_language.py
```diff
@@ -0,0 +1,23 @@
...
+
+active_feeds = [feed for feed in all_italian_feeds if not feed.deactivated]
+inactive_feeds = [feed for feed in all_italian_feeds if feed.deactivated]
+
+print(f"Feeds for {language.name}: ")
+for feed in active_feeds:
+    deactivated_status = "Deactivated" if feed.deactivated else ""
+    print(f"  - {feed.title} ({feed.url.domain_name()})")
```
tfnribeiro, line 17: I think it might be better to have the full url, rather than the domain name otherwise we will end up with situations like:

```
- Indland | DR (http://www.dr.dk)
  - Udland | DR (http://www.dr.dk)
  - Penge | DR (http://www.dr.dk)
  - Politik | DR (http://www.dr.dk)
  - Viden | DR (https://www.dr.dk)
  - Sport | DR (http://www.dr.dk)
  - Vejret | DR (http://www.dr.dk)
```
One of the uses for this tool could be to check if we have a wrong RSS feed, that is no longer available. Or that we don't have the right feed handler, so maybe adding the feed handler could be a good idea?

tfnribeiro (changes requested):

tools/feeds_for_language.py
mircealungu: let's do that!

mircealungu:
tak @tfnribeiro . addressed your changes. tool is good to be used for now.

PR #312, opened 2024-12-12
New exercise scheduling removing a bit of duplication and possibly fixing bug

mircealungu:
Can you have a look at this @tfnribeiro ?


tfnribeiro (approved):
I think this looks good - It makes sense to move the `there_was_no_need_for_practice_today` to the parent class rather than defining it in both children. 

Let's merge this, I will have to double check that all cycles are how we expect.

mircealungu:
Yes @tfnribeiro - good point with moving the function to the parent component!

PR #313, opened 2024-12-14
Commitment Feature - Research Project (Johanna and Chloe)

chloeolson:
Includes all api handling and calls, UserCommitment class, and testing instructions and scripts.


zeeguu/api/endpoints/user_activity_and_commitment.py
```diff
@@ -0,0 +1,109 @@
...
+def user_commitment_info():
+    """
+    Updates the number of days and minutes the user wants to practice
+    under settings
+    """
+    user = User.find_by_id(flask.g.user_id)
+    data = flask.request.form
+    user_minutes = data.get("user_minutes")
+    user_days = data.get("user_days")
+    commitment = UserCommitment.query.filter_by(user_id=user.id).first()
```
tfnribeiro, line 105: This should be done in the Class, rather than in the endpoint.

zeeguu/api/endpoints/user_activity_and_commitment.py
```diff
@@ -0,0 +1,109 @@
...
+    commitment.user_minutes = int(user_minutes)
+    commitment.user_days = int(user_days)
+    zeeguu.core.model.db.session.commit()
```
tfnribeiro, line 108: Also, you didn't add the object to the session - usually the pattern is:

1. Update the object - this is done at the class, returning the updated object.
2. You can then add that object to the session and commit.

You can see the discussion for this here: https://github.com/zeeguu/api/discussions/210

zeeguu/api/endpoints/user_statistics.py
tfnribeiro, line 1: These comments seem like personal notes, I'd also remove them.

zeeguu/core/model/user_commitment.py
tfnribeiro, line 1: As I stated in the endpoint code, most of the logic should be moved to the class rather than the endpoint you can take a look at other model files for inspiration, for example `TopicSubscription` doesn't have much logic.

zeeguu/core/user_statistics/activity.py
tfnribeiro, line 1: These comments seem like personal notes, I'd also remove them.

tfnribeiro (changes requested):
Overall, most of the logic is good, but it needs to be moved around so it's easier to maintain going forward. 

It also makes sense to add testing to the features you added, rather than suggesting a manual testing of the user as it's not sustainable to do every time there is an update. Ideally, you would add a test case for testing if a user is on streak or not for two base examples. You can have a look at the code under `zeeguu/core/test` where you have different tests for different features. This is what we use to ensure we don't break anything when deploying new features.

Other small things are some development comments and files that shouldn't be part of the PR.

docker-compose.yml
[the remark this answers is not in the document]
mircealungu: Is there such a thing? in PyCharm I have syntax highlighting but not auto-formatting .

tools/migrations/24-12-13--adding-user-commitment-table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE user_commitment (
+    user_id INT PRIMARY KEY,
+    user_minutes INT,
```
mircealungu, line 3: there is no reason to prefix the `minutes`, `days` with `user`. we already have the info that this table pertains to a user in the table name.

zeeguu/api/endpoints/user_activity_and_commitment.py
```diff
@@ -0,0 +1,109 @@
...
+@requires_session
+def user_activity_and_commitment():
+    """
+    Returns all the user activity and commitment data 
+    """
+    user = User.find_by_id(flask.g.user_id)
+    user_commitment = UserCommitment.query.filter_by(user_id=user.id).first()
+    user_activities = activity_duration_by_day(user)
+    return json_result({
+        "user_minutes": user_commitment.user_minutes,
```
mircealungu, line 26: `user_minutes` is not informative enough. `committed_minutes` or `commitment_minutes` is better.

zeeguu/api/endpoints/user_activity_and_commitment.py
```diff
@@ -0,0 +1,109 @@
...
+@requires_session
+def user_commitment():
+    """
+    Returns all user commitment data
+    """
+    user = User.find_by_id(flask.g.user_id)
+    user_commitment = UserCommitment.query.filter_by(user_id=user.id).first()
+    print(user_commitment.user_minutes)
+    return json_result({
+        "user_minutes": user_commitment.user_minutes,
```
mircealungu, line 45: ditto

zeeguu/api/endpoints/user_activity_and_commitment.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+# When the user meets their weekly goal, this method updates the consecutive weeks count in database
+@api.route("/user_commitment_update", methods=["POST"])
+@cross_domain
+@requires_session
+def user_commitment_update():
+    user = User.find_by_id(flask.g.user_id)
+    data = flask.request.form
+    consecutives_weeks = data.get("commitment_and_activity_data")
```
mircealungu, line 85: Why is this mismatch between the name of the local variable and the name of the data that's in the POST variable? 

It is suspicious to me that the web frontend would send back to the server information about consecutive weeks. all the necessary information for counting the consecutive weeks is on the server already. why does the react app have to send this?

zeeguu/api/endpoints/user_activity_and_commitment.py
mircealungu: Yes. In fact, you could even do it in the `User` class, so you can write `user.getCommitment()`. 

Which brings me to, the following questions:
- what happens for users who do not have a commitment? should there be conditions for them?
- UI-wise, should there be an option for the user who creates an account to "Skip" making a commitment. grayed out, not encouraged, but still present there?

mircealungu (approved):
I just had a look at this.
I agree with the observations of Tiago and added a few observations of my own.

docker-compose.yml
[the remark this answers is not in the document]
tfnribeiro: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml This seems to provide auto-formatting.

mircealungu:
closing this for now. it's too old - you'll open a new one when you're ready.

PR #317, opened 2025-01-17
Tokenization + Token 'Coordinates'

tfnribeiro:
With this change, we aim to have the backend tokenize all the texts that are sent to the frontend, with the main goal of allowing us to better tokenize the text and to facilitate the ability to show previous interactions on the text for our users. It should also improve our context extractor by better sentence detection. 

## Changes

- Introduced new columns to the **text** and **bookmark** table. These fields contain explanation on their respective model files. 
- Added a script to add the coordinates for old bookmarks. This script works for about 89% of bookmarks, I didn't look much into the errors, but it seems that it has to do with some small tokenization issues between the word and the sentence. Other cases, we didn't store a link to the original articles. If we want to increase this rate, I can spend a bit more time on it, but this feature is expected to work once the frontend and backend are aligned.
- Added a small script to allow for testing the tokenizer with a specific language.
- Updated the endpoints to receive and send the coordinates of the bookmarks.
- Updated the endpoints to allow the frontend to tokenize texts.
- `mp3_of_full_article`, used the contents stored in the DB rather than sending them from the frontend. 
- Added a util function to parse booleans from the frontend (they come as strings)
- Extend the utils/text to provide tokens utilizing the NLTK tokenizers.
    - Created a Token class that holds the text and different properties of the token.
    - Created regular expressions to detect if the Token is an email, url or number

There were some cases that are handled based on the rendering in the frontend - I tried to leave comments where relevant to make it easier to read. 

## To-Do

- Add test cases to the different languages to test the tokenizer


tools/migrations/25-01-06--add_index_pointers.sql
```diff
@@ -0,0 +1,35 @@
...
+    COLUMN `in_content` TINYINT NULL
+AFTER
+    `token_i`;
+
+/* Bookmark update */
+ALTER TABLE
+    `zeeguu_test`.`bookmark`
+ADD
+    COLUMN `sentence_i` INT NULL
+AFTER
```
mircealungu, line 26: there is no `learned` column. should it be `learned_time`?

tools/migrations/25-01-06--add_index_pointers.sql
```diff
@@ -0,0 +1,35 @@
...
+ADD
+    COLUMN `paragraph_i` INT NULL
+AFTER
+    `article_id`,
+ADD
+    COLUMN `sentence_i` INT NULL
+AFTER
+    `paragraph_i`,
+ADD
+    COLUMN `token_i` INT NULL
```
mircealungu, line 13: why do we need to have a `token_i` inside of the text? Is it not sufficient to anchor the text at a given `paragraph_i` and  `sentence_i`? Then the bookmark is going to be anchored in that sentence at a given `token_i`?

mircealungu (changes requested):
a few questions about the new data model

tools/tokenizer_playground.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+app = create_app()
+app.app_context().push()
+
+while True:
+    clear_terminal()
+    print("Welcome to the Zeeguu Tokenizer")
+    language = Language.find(input("Enter language code (ex. 'en'): ").lower())
+    text = input("Enter text: ")
```
mircealungu, line 20: thanks for the playground.

it's interesting to see that a double quote (") at the beginning of a sentence is not considered `left_punct`, but a chevron («) is. But maybe it's not that surprising: how is the poor parser to know that this is a left or a right token when both have the same shape. 

Though you'd expect that “...” would fare better, though it doesn't: 

```
Got the following text:  “...”
“
{'is_left_punct': False,
 'is_like_email': False,
 'is_like_num': False,
 'is_like_url': False,
 'is_punct': True,
 'is_right_punct': True,
 'is_sent_start': True,
 'paragraph_i': 0,
 'sent_i': 0,
 'text': '“',
 'token_i': 0}
```

zeeguu/core/util/text.py
```diff
@@ -6,19 +6,247 @@
...
+        "spanish",
+        "swedish",
+        "turkish",
+    ]
+)
+
+
+class Token:
+    PUNCTUATION = "»«" + punctuation + "–—“‘”“’„¿"
+    LEFT_PUNCTUATION = "«<({#„¿["
```
mircealungu, line 45: image

in english the shape of left / right position aware double quotations is at the top: the 66 and 99 shapes. from: https://en.wikipedia.org/wiki/Quotation_mark#Summary_table

mircealungu (changes requested):
a few more questions.

tools/migrations/25-01-06--add_index_pointers.sql
tfnribeiro: I think it's better to make sure we anchor the text with all this information, as right now we are not guaranteed that we always start at a beginning of a sentence. I think it makes it more versatile. The token could also be part of another sentence than the one the text starts in.

zeeguu/core/util/text.py
```diff
@@ -6,19 +6,247 @@
...
+        "slovene",
+        "spanish",
+        "swedish",
+        "turkish",
+    ]
+)
+
+
+class Token:
+    PUNCTUATION = "»«" + punctuation + "–—“‘”“’„¿"
```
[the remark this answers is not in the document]
tfnribeiro: Aren't they part of the punctuation string? `!”#$%&'()*+,-./:;<=>?@[\]^_`{|}~`

zeeguu/core/util/text.py
```diff
@@ -6,19 +6,247 @@
...
+            "sent_i": self.sent_i,
+            "token_i": self.token_i,
+            "paragraph_i": self.par_i,
+            "is_like_email": self.is_like_email,
+            "is_like_url": self.is_like_url,
+        }
+
+
+def split_into_paragraphs(text):
+    paragraph_delimiter = re.compile(r"\n\n")
```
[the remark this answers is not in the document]
tfnribeiro: I believe this is already handled by the first pass of readability.

tfnribeiro:
Currently the changes have only been tested against the Web Reader, not the extension.

I will create a branch for the extension to test this and upgrade to be in-sync with the tokenizer.

zeeguu/api/endpoints/bookmarks_and_words.py
```diff
@@ -134,19 +140,31 @@ def bookmarks_for_article(article_id, user_id):
     return json_result(dict(bookmarks=bookmarks, article_title=article.title))
 
 
-@api.route("/bookmarks_to_study_for_article/<int:article_id>", methods=["POST", "GET"])
+@api.route(
+    "/bookmarks_to_study_for_article/<int:article_id>",
+    methods=["POST", "GET"],
+)
+@api.route(
+    "/bookmarks_to_study_for_article/<int:article_id>/<with_tokens>",
```
tfnribeiro, line 148: Make post method

zeeguu/api/endpoints/exercises.py
```diff
@@ -22,52 +28,80 @@ def scheduled_bookmarks_to_study(bookmark_count):
...
 
 
-@api.route("/top_bookmarks_to_study", methods=["GET"])
+@api.route("/top_bookmarks_count", methods=["GET"])
 @cross_domain
 @requires_session
-def top_bookmarks_to_study():
+def top_bookmarks_count():
     """
     Return all the possible bookmarks a user has to study ordered by
```
tfnribeiro, line 44: Update comment

zeeguu/core/model/bookmark.py
```diff
@@ -103,10 +121,19 @@ def __repr__(self):
         )
 
     def serializable_dictionary(self):
```
tfnribeiro, line 123: as_dictionary,

Combine the two, with_exercise_info, with_tokens, ... (Check title if used).

zeeguu/core/model/bookmark.py
```diff
@@ -316,7 +362,17 @@ def find_or_create(
 
         article = Article.query.filter_by(id=article_id).one()
 
-        context = Text.find_or_create(session, _context, origin_lang, None, article)
+        context = Text.find_or_create(
```
tfnribeiro, line 365: Update text, so when a bookmark is updated we check if the text still has any rows pointing to it.

zeeguu/core/test/test_tokenizer.py
```diff
@@ -0,0 +1,294 @@
...
+            "techno-industriel",
+            "»",
+        ] == [t.text for t in tokens]
+        assert all([t.token_i == i for i, t in enumerate(tokens)])
+
+    def test_spanish_tokenization_1(self):
+        text = """¿qué es esto?"""
+        tokens = self.es_tokenizer.tokenize_text(text, False)
+        assert ["¿", "qué", "es", "esto", "?"] == [t.text for t in tokens]
+        assert all([t.token_i == i for i, t in enumerate(tokens)])
```
tfnribeiro, line 142: Explain why this function is here (use a function)

zeeguu/core/test/test_tokenizer.py
```diff
@@ -0,0 +1,294 @@
...
+            "Frost",
+            "geschaffenen",
+            "Serie",
+            "„",
+            "Twin",
+            "Peaks",
+            "“",
+        ] == [t.text for t in tokens]
+        assert all([t.token_i == i for i, t in enumerate(tokens)])
+        # assert tokens[10].is_left_punct
```
tfnribeiro, line 190: Remove these lines

zeeguu/core/test/test_tokenizer.py
```diff
@@ -0,0 +1,294 @@
...
+        ] == [t.text for t in tokens]
+        assert all([t.token_i == i for i, t in enumerate(tokens)])
+
+    def test_danish_tokenization_1(self):
+        text = """»Vi kan gøre det,« siger Mikkel."""
+        tokens = self.da_tokenizer.tokenize_text(text, False)
+        assert ["»", "Vi", "kan", "gøre", "det", ",", "«", "siger", "Mikkel", "."] == [
+            t.text for t in tokens
+        ]
+        # Question? should it be like this?
```
tfnribeiro, line 217: Remove

zeeguu/core/test/test_tokenizer.py
```diff
@@ -0,0 +1,294 @@
...
+            "på",
+            "Danske",
+            "Filmkritikeres",
+            "generalforsamling",
+            ".",
+        ] == [t.text for t in tokens]
+        assert tokens[1].is_like_num
+        assert all([t.token_i == i for i, t in enumerate(tokens)])
+
+    def test_danish_tokenization_2(self):
```
tfnribeiro, line 241: Check name of method

zeeguu/core/test/test_tokenizer.py
```diff
@@ -0,0 +1,294 @@
...
+        # Generated URLs from https://www.randomlists.com/urls?qty=50
+
+        emails = TokenizationTest._load_example_file("random_emails.txt")
+        tokens = [self.en_tokenizer.tokenize_text(e, False)[0] for e in emails]
+        for i in range(len(tokens)):
+            assert tokens[i].text == emails[i]
+            assert tokens[i].is_like_email
+
+    def test_number_detection(self):
+        # Generated URLs from https://www.randomlists.com/urls?qty=50
```
tfnribeiro, line 288: remove this comment

zeeguu/core/tokenization/tokenizer.py
```diff
@@ -0,0 +1,307 @@
...
+        "swedish",
+        "turkish",
+    ]
+)
+
+# Compile all regex before hand to speed up matching.
+
+PARAGRAPH_DELIMITER = re.compile(r"((\s?)+\\n+)")
+APOSTROPHE_BEFORE_WORD = re.compile(r" (')([\w]+)")
+PARTICLE_APOSTROPHE_BEFORE_WORD = re.compile(r"(\w+('|’))")
```
tfnribeiro, line 42: Give examples

zeeguu/core/tokenization/tokenizer.py
```diff
@@ -0,0 +1,307 @@
...
+        doc = self.nlp_pipeline(text)
+        current_paragraph = []
+        s_i = 0
+        for sentence in doc.sentences:
+            sent_list = []
+            is_new_paragraph = False
+            accumulator = ""
+            t_i = 0
+            for token in sentence.tokens:
+                t_dict = token.to_dict()
```
tfnribeiro, line 232: Call it a list, rather than dictionary

mircealungu (approved):
Amazing work! Let's merge!

mircealungu:
it does seem that you've forgot to upload the `test_data` files @tfnribeiro 👍 
image

tfnribeiro:
I did add the files:

image

mircealungu:
> I did add the files:
> 
> image

But i don't see it here on GH. 

image

Neither when I clone the repo. So what could be the explanation? 

image

PR #319, opened 2025-01-23
taking learning cycle into account when migrating from learnincycles …

mircealungu:
…to levels. i guess we have to do the same on the frontend for now.

### DIFF
--- a/zeeguu/core/word_scheduling/basicSR/four_levels_per_word.py
+++ b/zeeguu/core/word_scheduling/basicSR/four_levels_per_word.py
@@ -8,9 +8,9 @@ MAX_LEVEL = 4
 COOLING_INTERVAL_TO_LEVEL_MAPPING = {
     0: 1,
     ONE_DAY: 1,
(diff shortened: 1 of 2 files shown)


zeeguu/core/word_scheduling/basicSR/four_levels_per_word.py
```diff
@@ -8,9 +8,9 @@
 COOLING_INTERVAL_TO_LEVEL_MAPPING = {
     0: 1,
     ONE_DAY: 1,
-    2 * ONE_DAY: 2,
-    4 * ONE_DAY: 3,
-    8 * ONE_DAY: 4,
+    2 * ONE_DAY: 1,
```
merleschoen, line 11: I think this is a pretty good solution for bookmarks already in productive knowledge, but receptive bookmarks with cooling interval 8 would have to be answered  correctly 17 times in total. Is that what we want? Realistically speaking, most of our users won't have any productive bookmarks, so it might be worthwhile focusing on receptive bookmarks.

PR #321, opened 2025-01-30
FK Updates 

tfnribeiro:
- FK uses Token Length / Syllables and Sentence length to compute the difficulty of a text. We will have cases where we don't have a formula for new languages. To make better "guesses" what is the best formula, we can use the script to find which language is closest given these metrics. TO DO, the weights are slightly different, so maybe those should also be reflected in the distance calculation.
- Fixes a value in the Russian Formula, and Polish now uses the Russian values rather than german.
- Portugal now uses the Spanish weights rather than the English ones.

## Portugal FK (only 788 articles)

| English Constants | Spanish Constants |
| :-: | :-: |
| image | image |

### DIFF
--- a/tools/get_lang_stats.py
+++ b/tools/get_lang_stats.py
@@ -1,0 +1,130 @@
+# coding=utf-8
+
+import zeeguu.core
+from zeeguu.core.model import Article, Language
+import stanza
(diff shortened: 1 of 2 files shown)


mircealungu (approved):
absolutely super-cool!

PR #326, opened 2025-02-10
Fixes to Stanza model

tfnribeiro:
- Fixed path to be reachable
- In case the coordinates of the paragraph are null then use 0.


mircealungu (approved):
looks good!

PR #332, opened 2025-02-14
Update to tokenization

tfnribeiro:
- Found a case in French where the assumption that the apostrophe wouldn't be followed by a single character doesn't hold. Made that condition specifically for punctuation rather than a token in general.

"n'a" - should be a token, rather than "n'", "a".

### DIFF
--- a/zeeguu/core/tokenization/stanza_tokenizer.py
+++ b/zeeguu/core/tokenization/stanza_tokenizer.py
@@ -114,8 +114,8 @@ class StanzaTokenizer(ZeeguuTokenizer):
                     particle_with_apostrophe
                     and particle_with_apostrophe.group(0) == text
                     and i + 1 < len(sentence.tokens)
-                    and len(sentence.tokens[i + 1].text)
-                    > 1  # avoid situations like call'? where it's followed by a punctuation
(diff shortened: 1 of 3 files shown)


tfnribeiro:
Note, this will affect at least the French coordinates from the script (since they will be positioned -1, if they occur after cases such as n'a)

tfnribeiro:
Found that is_like_symbols was using the wrong condition.

zeeguu/core/tokenization/stanza_tokenizer.py
```diff
@@ -114,8 +114,8 @@ def tokenize_text(
                     particle_with_apostrophe
                     and particle_with_apostrophe.group(0) == text
                     and i + 1 < len(sentence.tokens)
-                    and len(sentence.tokens[i + 1].text)
-                    > 1  # avoid situations like call'? where it's followed by a punctuation
+                    and sentence.tokens[i + 1].text
+                    not in Token.PUNCTUATION  # avoid situations like call'? where it's followed by a punctuation
```
mircealungu, line 118: next thing we find out, in some polish exceptions you can have `l'-brzvhj` or something :)))

mircealungu (approved):
cool find!

PR #337, opened 2025-03-03
Update ReadMe (container specification)

martinkimoe:
"docker compose up" runs all containers, which breaks as there are multiple instances of zapi. It should therefore be "docker compose up zapi_dev_translations".


tfnribeiro (approved):
Indeed!

tools/migrations/25-02-28--add_video-tables.sql
```diff
@@ -0,0 +1,54 @@
+-- Step 1: Create the `yt_channel` table (assuming 'language' table exists)
+CREATE TABLE `yt_channel` (
```
mircealungu, line 2: this should be in a different PR. This one is called Update Readme :)
Can you do that?

mircealungu (changes requested):
oen main comment.

tools/migrations/25-02-28--add_video-tables.sql
martinkimoe: Yes. I created a new branch with only that commit - #341

PR #343, opened 2025-03-11
Fix for extension not saving prgoress

tfnribeiro:
- We noted that especially in smaller articles our estimation based on the scrolling percentage can perform quite poorly (not showing small articles as read). Instead of using the scroll percentage, we now use the total duration the user has read the article. This was made to ensure that if the user is off by a few seconds opening the article can resolve that issue.

### DIFF
--- a/zeeguu/core/model/user_reading_session.py
+++ b/zeeguu/core/model/user_reading_session.py
@@ -146,12 +146,19 @@ class UserReadingSession(db.Model):
 
     @classmethod
     def get_total_reading_for_user_article(cls, article, user):
-        return (
-            cls.query(sum(cls.duration))
-            .filter(cls.article == article)
-            .filter(cls.user == user)
-            .one()
-        )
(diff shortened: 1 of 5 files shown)


zeeguu/core/util/time_conversion.py
```diff
@@ -0,0 +1,5 @@
+def ms_to_m(ms: int):
```
mircealungu, line 1: love this file and function :)

zeeguu/core/model/user_reading_session.py
```diff
@@ -147,6 +144,22 @@ def find_by_cohort(
         sessions = query.all()
         return sessions
 
+    @classmethod
+    def get_total_reading_for_user_article(cls, article, user):
```
mircealungu, line 148: looks good.

mircealungu (approved):
looks good!

PR #345, opened 2025-03-12
Data model update

tfnribeiro:
- Added the new tables to the Database to support the Source 'parent' class for all types of content in the application.
- Added columns to allow mappings of the existing objects to the new model.
- Updated the language distance tool to normalized the data across the languages before calculating distance.
- Updated tests to fit the new model


zeeguu/core/model/bookmark.py
```diff
@@ -419,18 +528,27 @@ def find_or_create(
...
                 token_i=token_i,
                 total_tokens=total_tokens,
+                context=context,
                 level=level,
             )
         except Exception as e:
             raise e
 
         session.add(bookmark)
+        if context_type is not None:
```
mircealungu, line 545: Could this method take the Contextidentifier as parameter?

mircealungu (changes requested):

mircealungu (approved):
amazing work!

PR #346, opened 2025-03-19
Ml topic classification based on text

tfnribeiro:
- Updated docker compose to use the Zeeguu data + the Zeeguu resources to download the models/store ES data.
- Simplified the pipeline to be more generic and easier to use. This is to easily support the fact that we are going to have multiple sources in ES and not just articles.
- Now when calling the `build_elastic_semantic_sim_query_for_topic_cls`, we pass a text + ids we do not want to be used.
- I also simplified the call to get the Topic Object by creating the method  `get_topic_classification_based_on_similar_content`. This makes it more easily to re-use the same logic used for article inferance.
- Created a simple demo where a text is classified using the KNN classifier.
- Increased the number of candidates when retrieving the neighbours.

@mircealungu I realized I reverted this change:

```
 ports:
      - "127.0.0.1:8080:9001"
```

Did we decide if we wanted to use the mapping or update the configuration for the api.config?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,4 @@
 services:
-
   # main db
   mysql:
     image: mysql:5.7
@@ -13,7 +12,6 @@ services:
     networks:
       - zeeguu_backend
     restart: unless-stopped
-
 
   # main db
   fmd_mysql:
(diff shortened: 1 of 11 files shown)


mircealungu (approved):
nice stuff!

PR #348, opened 2025-03-20
Add orphaned context type

tfnribeiro:
- Text doesn't have and FK key so there could be instances where the id doesn't have a corresponding article.
- These with be classified as OrphanContext and in the future we can update any linking bookmarks to content we delete as this type.


mircealungu (approved):
nice. let's do this!

PR #349, opened 2025-03-24
Fixed Source Broken Update

tfnribeiro:
- Fix error spotted in Sentry.

This led me to find an interesting point we should consider. Now, spammy articles that have a text we have encountered before are all linked to the same source, and there might be a case where in the past, they weren't considered as broken, but now they are. This will make it easier to spot these ocurrences.


mircealungu (approved):
tak!

PR #350, opened 2025-03-25
Fix migration script

tfnribeiro:
- Since the query was based off source_id being none, the offset was "skipping" articles as the starting article will always be the first article in the DB without a source id.
- Added an extra condition when finding a source that isn't broken but it's being attempted to be found with a broken flag. If this is the case, then we make it broken.

### DIFF
--- a/zeeguu/core/model/source.py
+++ b/zeeguu/core/model/source.py
@@ -46,9 +46,9 @@ class Source(db.Model):
         return self.source_text.content
 
     @classmethod
-    def find(cls, id: int):
+    def find_by_id(cls, id: int):
         try:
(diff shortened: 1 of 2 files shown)


zeeguu/core/model/source.py
```diff
@@ -64,10 +64,22 @@ def find_or_create(
     ):
         source_text = SourceText.find_or_create(session, text, commit=commit)
         try:
-            return cls.query.filter_by(
+            source = cls.query.filter_by(
                 source_text=source_text, source_type=source_type, language=language
             ).one()
 
+            if source.broken == 0 and broken > 0:
```
mircealungu, line 71: 👍

mircealungu (approved):
:+1

PR #351, opened 2025-03-27
Updated backend to receive web info

tfnribeiro:
Update to allow Web to send information to write to the new data model. We add a new source_id table to the user_activity_data to track the interactions of the users based on source_id.

-  (mysql_to_elastic) Added a comment line to make it easier to uncomment and run the script.
- Created a helper method to extract the context_identifier from the frontend.
- Adapted tests to allow for SourceRule and BookmarkContextRule
- Removed unnecessary test: test_contribute_own_translation_no_context_type
- Bookmark has been re-worked to use the new classes. A new method was added to get the title (which is given based on it's context.) 
- Simplified user_article_source to extract the source_id in case an article_id is sent. This is done since the frontend can send either article ID or source_id (what if they pass both)?


mircealungu (approved):
nice work!

mircealungu (approved):
looks good!

PR #356, opened 2025-04-03
Available Languages reports languages crawled.

tfnribeiro:
- Changed available languages call to show the languages we are actively crawling in the backend.


mircealungu (approved):
tak @tfnribeiro !

PR #359, opened 2025-04-04
Add code to update existing articles id to ES

tfnribeiro:
- Created a update_elastic_ids.py which updates all documents in ES to have a field: article_id
- In the future, we will instead query using the term, rather than the ES id (the ES id will be unique and disconnected from either source or article_id).

After this change, we will have videos which will share some properties, but then will have a video_id field.


mircealungu (approved):
looks good

PR #364, opened 2025-04-15
configuring the zapi with YOUTUBE_COOKIES_FILE_PATH

mircealungu:
so this should work you say folks? @martinkimoe @silashoeyer ?


mircealungu:
not needed anymore; we're trying out a new library

PR #365, opened 2025-04-16
fix-caption-downloader

silashoeyer:
Change caption downloader to youtube-transcript-api instead of yt-dlp. This should not require login, and it seems to be faster as well.
https://github.com/jdepoix/youtube-transcript-api

TODO:
- Make sure all tables where text is stored have characterset utf8mb4 and collate utf8mb4_bin (I think this was done Monday)
- If this is working, the video_downloader.py needs to uncomment the rest of the topics. And the video_crawler should set the number of videos to 50.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -68,7 +68,5 @@ pydantic>=1.10.8
 stanza==1.10.1
 
 # For videos
-webvtt-py
-yt-dlp
-ffmpeg
 emoji
+youtube_transcript_api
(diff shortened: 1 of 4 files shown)


silashoeyer:
VTT column is also removed since youtube-transcript-api doesnt download captions as vtt, but as caption dictionaries.

mircealungu (approved):
looks simpler. i like this. (i only removed some commented out code).
let's deploy and test it on the server.

PR #366, opened 2025-04-18
Transfer captions to production via json file

martinkimoe:
This PR includes to main changes:

**get_captions() in youtube_api.py:** Here we commented out the code that uses the youtube_transcript_api to fetch captions and inserted code that tries to find the captions in a manually added captions.json.

**captions.json**: We have run the crawler on our own computer with "crawl(50)" and the the topic "Sport" and category "Entertainment". This should guarantee that at least some of the videos that the crawler find, will have their captions in this file.

**add_captions_from_file.py**: Run this file to add captions from captions.json to already crawled videos (with broken status=1)

If this succeeds, then we can use this as a temporary way of crawling videos. We just have to crawl on our own computers first and then push the captions to production via captions.json. And then run the crawler with the same categories/topics on the production server.


zeeguu/core/youtube_api/youtube_api.py
```diff
@@ -162,49 +163,93 @@ def _get_thumbnail(item):
 
 
 def get_captions(video_unique_key, lang):
+    # try:
```
mircealungu, line 166: it is confusing if we keep commented out code in version control. either remove it, or maybe move it to it's own function (e.g. `get_captions_with_yttapi`).

mircealungu: Alternatively, move the current implementation to `get_captions_from_json`.

zeeguu/core/youtube_api/captions.json
mircealungu, line 1: we only version control code. this file does not belong in git. it will just increase the size of the GH repo in perpetuity. better remove it from the PR, and send it to me separately via mail, or teams

martinkimoe: Got ya. Then how do we reference to the file in the code?

zeeguu/core/youtube_api/youtube_api.py
martinkimoe: Done in commit: c576911

PR #374, opened 2025-05-02
Fix video language detection

silashoeyer:
- Fix the bug where videos in the wrong language are not set as broken. The problem was that we didn't return after finding the video broken. Before we would still try to find captions for videos in the target language, and sometimes they would be available. Now we save the video as broken as soon as the wrong language is detected.

- Fix another bug where some videos dont have a duration value.

- Clean up some duplicated code.


silashoeyer:
This will not remove videos that are currently wrong. But it will work on future videos.

PR #377, opened 2025-05-13
Scheduler using the "Max Words In Learning" user preference 

mircealungu:
Cleaning up some of the endpoints for bookmark scheduling. 

Added support for a MAX_WORDS_TO_SCHEDULE preference and updated the scheduling algo accordingly.


mircealungu:
One question for you @tfnribeiro. 

I defined a default value for this preference to be 20. 

As a migration, I have a script that will delete schedule words and leave the most advanced 20 in all the user accounts. I remember we looked at this together and we agreed that this would not affect many users, right? Or did we agree on a larger number of studied words?

PR #379, opened 2025-05-21
Added the Meaning class as a first step towards scheduling meanings not bookmarks

mircealungu:
This because one can have multiple bookmarks that point to the same meaning. It is silly to schedule two bookmarks with the same meaning just because they are different.


mircealungu:
self-reviewed. let's deploy!

PR #390, opened 2025-06-17
fixing the own article upload bug found by Tiago

mircealungu:
Fixing https://github.com/zeeguu/web/issues/798 opened by @tfnribeiro .

The `upload_own_article` and `update_own_article` were not migrated to work with the new Source table. Now it should all work hunky dory.


tfnribeiro:
I thought it would be something related to this, thanks for fixing it.

Later I will try a sample upload and then re-adding the texts - Good thing I didn't manually type everything because it was a total of 3 pages!

mircealungu:
> I thought it would be something related to this, thanks for fixing it.
> 
> Later I will try a sample upload and then re-adding the texts - Good thing I didn't manually type everything because it was a total of 3 pages!

eheh. you're too experienced to not type straight into a textbox on the web :) 

i've merged and deployed. give it another try at some point.

PR #516, opened 2026-04-01
Feature/leaderboards

gabortodor:


tools/generate_fake_articles.py
xXPinkmagicXx: Do we still want these scripts to generate test data?

Issue #14, opened 2018-03-20
remove the dashboard_enabled condition 


mircealungu:
it's ugly and the effect of deactivating the dashboard
can be obtained if we undefine the envvar. 

in fact, we could also load the dashboard from an `app.config['dashboard_config']`

https://github.com/zeeguu-ecosystem/Zeeguu-API/blob/a8dd603478b25dad1f622abb1c2bedd400758b1a/zeeguu_api/app.py#L38

PR #544, opened 2026-04-18
Verbal flashcards API implementation

jakso535:
## Why this Pull Request exists
This PR adds the backend required for a new spoken vocabulary exercise called **Verbal Flashcards**.

The new backend work does four jobs:
1. expose verbal flashcards as a new exercise source,
2. transcribe uploaded learner audio,
3. evaluate the transcription against the expected answer with learner-friendly tolerance,
4. store the final outcome inside Zeeguu's normal exercise and scheduling system.

The result is not just “speech-to-text attached to flashcards.” It is a full backend workflow for spoken practice.

---

## High-level architecture
The implementation is split into three parts:

### 1. Main Zeeguu API endpoint layer
This is the part that serves flashcards, receives answers, and integrates results into the existing scheduling model.

### 2. ASR service client layer
This decides which ASR worker should handle a transcription request based on language.

### 3. Dedicated ASR worker microservice
Each worker loads a single speech-recognition model and exposes transcription over HTTP.

This separation is a strong design choice because it keeps the core application free from model-specific dependencies and makes multilingual expansion much easier.

---

### 1. A new verbal flashcards endpoint module
The main new backend module introduces the feature's business logic, API routes, matching logic, and persistence behavior.

### 2. Feature-flag support
The feature is gated behind a dedicated feature toggle so it can be rolled out safely.

### 3. An ASR client for routing by language
The API can forward audio to different workers depending on the learner's studied language.

### 4. A dedicated ASR worker service
A small Flask microservice loads one ASR model and handles transcription requests.

### 5. Gunicorn config for the worker
The worker is packaged with deployment settings for production-style serving.

### 6. Tests
The feature includes focused tests for:
- endpoint behavior,
- feature gating,
- flashcard selection,
- fuzzy matching,
- worker routing,
- error handling,
- answer submission.

---

## `verbal_flashcards.py`
This is the main backend implementation.

#### Feature access control
The module checks whether the current user is allowed to use verbal flashcards. If not, it returns a feature-disabled response.

#### Flashcard generation
The function `get_flashcard_collection(user)` converts the learner's scheduled study words into spoken flashcards.

Important behavior:
- only words at **level 3 or above** are included,
- duplicate source words are removed,
- each returned card includes both prompt and expected answer fields,
- cards are derived from existing user words rather than from a separate flashcard database.

This is a smart choice because verbal practice is being layered onto Zeeguu's existing learning model instead of duplicating vocabulary state.

#### Schedule recovery for higher-level words
The helper `_ensure_schedule_for_verbal_flashcard(user_word)` makes sure a word has a schedule row even if it is not currently in the normal exercise pipeline.

This matters because verbal flashcards can target mature words. The feature therefore needs a way to write outcomes back into the existing scheduling system without resetting the learner's progress.

---

### `GET /verbal_flashcards`
Returns the learner's currently available verbal flashcards.

#### What it supports
- optional `limit`
- optional `offset`
- feature-gate enforcement
- pagination metadata in the response

#### Why it matters
This is the entry point for the frontend session. It turns scheduled Zeeguu study words into a frontend-ready spoken exercise list.

---

### `POST /verbal_flashcards/transcribe`
Accepts an uploaded audio file and returns a transcription.

#### Main behavior
- validates that a file exists,
- validates filename presence,
- checks whether the feature is enabled,
- reads the learner's learned language,
- sends the audio to the correct ASR worker,
- returns the transcribed text.

#### Error handling
It returns specific status codes for different failure modes:
- `400` for missing or invalid file input,
- `503` when no ASR worker is configured for the language,
- `502` when the worker request fails,
- `500` for unexpected internal errors.

---

### `POST /verbal_flashcards/check_pronunciation`
Accepts:
- `user_speech`
- `expected_text`

and returns structured pronunciation analysis **without storing progress**.

#### Why this endpoint is useful
It separates *evaluation* from *persistence*.

That gives the frontend a safe way to:
- show immediate feedback,
- decide whether the learner should retry,
- avoid writing partial attempts into the database too early.

This separation is one of the strongest architectural decisions in the PR.

---

### `POST /verbal_flashcards/submit`
Accepts the final attempt result and records it as an exercise outcome.

#### Expected payload includes
- `flashcard_id`
- `user_answer`
- `is_correct`
- `answer_source`
- `response_time_ms`
- optional `session_id`

#### Important behavior
- validates required fields,
- resolves the flashcard for the current user,
- coerces and validates timing/session data,
- recalculates accuracy when a user answer is present,
- allows fuzzy acceptance to override an initially false `is_correct`,
- ensures a schedule row exists,
- writes the result as a normal Zeeguu exercise outcome,
- tags the exercise source as `Verbal Flashcards`.

This endpoint is what makes the feature part of Zeeguu's real learning pipeline rather than just a standalone pronunciation demo.

---

## Matching and evaluation logic
ASR output is noisy, especially for foreign-language learners. If the system demanded exact string equality, the feature would feel unfair and would reject many answers that are close enough to be pedagogically useful.

This PR solves that problem by combining several layers of normalization and fuzzy matching.

### `sanitize_spoken_text`
Normalizes casing, punctuation, and spacing while preserving Danish characters.

### `canonical_danish_form`
Converts common alternative spellings into stable Danish written forms, such as:
- `aa` -> `å`
- `ae` -> `æ`
- `oe` -> `ø`

### `asr_tolerant_danish_form`
Applies even more permissive transformations for ASR comparison, such as:
- `æ` -> `e`
- `ø` -> `o`
- `å` -> `a`
- simplifying some initial and final consonant patterns.

This is the feature's “grace for language learners” layer.

---

### Damerau-Levenshtein distance
Used to measure edit distance with support for insertion, deletion, substitution, and transposition.

This is useful for small ASR spelling mistakes and near-miss transcriptions.

### Jaro and Jaro-Winkler similarity
Used to capture similarity in a more flexible way, especially for short strings.

### Boundary-aware Jaro-Winkler
The implementation also compares reversed strings so that dropped initial sounds are not punished too harshly.

### Length-aware acceptance threshold
Short words use stricter thresholds, while longer words can tolerate more variation.

That is a very sensible design choice because one-letter differences matter much more in very short words.

---

## Word-level accuracy analysis
The evaluation does not just compare full phrases.

Instead, for each expected word it:
- looks for the best unmatched spoken candidate,
- computes similarity scores,
- records whether the word matched,
- records whether it was exact, normalized, fuzzy, or close,
- builds a word-by-word result structure.

The response includes:
- overall accuracy,
- accepted accuracy,
- accepted word count,
- acceptance boolean,
- feedback label,
- detailed word matches.

This is what makes the frontend's word breakdown possible.

---

## `user_feature_toggles.py`
This file adds the feature toggle for `verbal_flashcards`.

### How the toggle works
The feature is enabled only for users whose stored invitation code appears in the environment variable:
- `VERBAL_FLASHCARDS_INVITE_CODES`

### Why this matters
This is a good rollout strategy for an experimental feature because it allows:
- controlled access,
- language-specific pilots,
- staged testing with selected cohorts or invite codes.

---

## `asr_service_client.py`
This file is the bridge between the main API and the ASR workers.

#### Parse worker mappings
It accepts configuration strings such as:
- `da=http://asr-da:5002`
- `de=http://asr-de:5002`
- `fr=http://asr-fr:5002`

and converts them into a language-to-URL map.

#### Resolve the correct worker
Given a learner language code, it selects the correct worker URL.

#### Fallback local development setup
If nothing is configured, it falls back to a local Danish worker mapping.

#### Perform the transcription request
It sends the audio file plus `language_code` to the selected worker and normalizes request failures into specific exceptions.

### Why this matters
This file is what turns the system from “one speech service” into a **language-aware worker architecture**.

---

## `app.py` (ASR worker microservice)
This file implements the dedicated ASR worker.

### Purpose
Each worker instance owns exactly one language model.

That means the main API does not run speech recognition directly. Instead, it forwards the request to the worker that owns the model for the learner's language.

### Main environment variables
- `ASR_LANGUAGE_CODE`
- `ASR_MODEL_NAME`
- `ASR_WORKER_NAME`
- `ASR_SERVICE_PORT`

### Model loading
At startup, the worker tries to import:
- `nemo.collections.asr`
- `pydub`

and then loads the configured model using `ASRModel.from_pretrained(...)`.

### Audio preprocessing
Before transcription, the worker:
- reads the uploaded audio,
- converts it to mono,
- resamples it to 16 kHz,
- exports it to a temporary WAV file,
- runs transcription on that normalized file.

This is important because speech models are much more reliable when audio is normalized to the format they expect.

#### `GET /health`
Returns worker status, language, name, and whether the model is loaded.

#### `POST /transcribe`
Accepts a file upload plus optional `language_code`, validates that the request matches the worker's configured language, and returns a transcription.

### Why this matters
This worker is small, focused, and operationally clear. It isolates model dependencies and makes future multilingual scaling straightforward.

---

## `gunicorn.conf.py`
This file provides deployment config for the ASR worker.

### What it sets
- bind address
- worker count
- thread count
- timeouts
- log output
- startup and ready hooks

### Why it matters
This turns the worker from a dev-only Flask app into something that is ready to run as a standalone service.

---

## `test_verbal_flashcards.py`
This file provides targeted coverage for the feature.

#### Feature availability
- users get `404` when the feature is disabled.

#### Flashcard generation
- only level-3+ words are returned,
- duplicate origin words are removed,
- pagination works.

#### Text normalization and matching
- Danish sanitization behaves correctly,
- canonical spelling normalization works,
- ASR-tolerant forms work,
- fuzzy matching accepts common learner/ASR variants,
- multi-word answers can still be accepted when word order differs.

#### Endpoint behavior
- pronunciation checking validates required fields,
- transcription endpoint returns expected data,
- missing worker configuration returns `503`,
- invalid session ids return `400`.

#### Answer submission
- exercise outcomes are recorded,
- fuzzy acceptance can override a false incoming correctness flag,
- invalid response times are coerced safely,
- non-eligible words return `404`.


requirements.txt
```diff
@@ -10,6 +10,7 @@ en_core_web_md @ https://github.com/explosion/spacy-models/releases/download/en_
 de_core_news_md @ https://github.com/explosion/spacy-models/releases/download/de_core_news_md-3.7.0/de_core_news_md-3.7.0-py3-none-any.whl
 da_core_news_md @ https://github.com/explosion/spacy-models/releases/download/da_core_news_md-3.7.0/da_core_news_md-3.7.0-py3-none-any.whl
 nltk
+nemo_toolkit[asr]
```
mircealungu: This pulls torch and several GB of model-tooling dependencies into the main API image. The dedicated `asr_service/` already lists `nemo_toolkit[asr]` in its own `requirements.txt` — that's where it belongs. The main API only proxies audio to the worker; it just needs `requests`. Please remove this line.

requirements.txt
```diff
@@ -20,7 +21,7 @@ matplotlib
 seaborn
 
 # Database
-mysqlclient
+mysqlclient==2.2.7
```
mircealungu: This pin is unrelated to verbal flashcards. Either split it into its own PR or add a line in the PR description explaining why it's needed.

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,117 @@
...
+import logging
+
+import requests
+from flask import has_app_context, current_app
+
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ASR_SERVICE_TIMEOUT = float(os.environ.get("ASR_SERVICE_TIMEOUT", "30"))
+LOCAL_DEV_ASR_SERVICE_URLS = "da=http://127.0.0.1:5002"
```
mircealungu: Falling back silently to `http://127.0.0.1:5002` when `ASR_SERVICE_URLS` is unset means a misconfigured production deploy quietly tries localhost and then fails with a 502 on connection refused. Prefer to only use this fallback in a dev context (e.g. when `FLASK_ENV == 'development'`) and raise `ASRServiceNotConfigured` otherwise.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,756 @@
+import traceback
```
mircealungu, line 1: This file is 756 lines and mixes four concerns: routes, flashcard-from-scheduled-words mapping, Danish text normalization, and fuzzy matching. Consider splitting into:

- `zeeguu/core/verbal_flashcards/text_normalization.py` (canonical + ASR-tolerant)
- `zeeguu/core/verbal_flashcards/fuzzy_match.py` (DL + JW + `score_word_match` + `calculate_accuracy`)
- `zeeguu/core/verbal_flashcards/flashcard_selection.py` (collection + `_ensure_schedule_for_verbal_flashcard`)

and keep this file as a thin route layer.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,756 @@
...
+    return schedule
+
+
+# ====================================
+# Helper Functions
+# ====================================
+FUZZY_ACCEPTANCE_BUFFER = 0.08
+
+
+def canonical_danish_form(word):
```
mircealungu: These normalizers are Danish-specific but live behind generic-sounding names — `sanitize_spoken_text`, `score_word_match`, `fuzzy_match_threshold` all implicitly assume Danish. When this is extended to `de`/`fr` you'll need a per-language normalizer registry. Worth setting up the abstraction now (e.g. `normalizer_for(language_code)`) even while only Danish is implemented, so future languages don't require refactoring every call site.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,756 @@
...
+    standard exercise pipeline. Create a schedule row without resetting the level
+    so the word appears in /words after it is practiced.
+    """
+    schedule = FourLevelsPerWord.find(user_word)
+    if schedule:
+        return schedule
+
+    schedule = FourLevelsPerWord(user_word=user_word)
+    schedule.next_practice_time = datetime.now()
+    schedule.consecutive_correct_answers = 0
```
mircealungu: This commits mid-request, and then `report_exercise_outcome` commits again a few lines later in the submit flow. Two commits per submit leaves a window where partial state lands on exception between them. Prefer `db_session.flush()` here and let the downstream `report_exercise_outcome` commit the aggregate.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,756 @@
...
+        return "Success"
+    return "Very close, try again"
+
+
+def transcribe_audio(audio_file, language_code=None):
+    """
+    Transcribe audio by routing the request to the dedicated ASR worker that
+    owns the model for the user's learned language.
+    """
+    audio_bytes = audio_file.read()
```
mircealungu: claude says: no file size limit on the upload — `audio_file.read()` reads the full body into memory unconditionally. A large or malicious POST can exhaust worker memory. Either set `MAX_CONTENT_LENGTH` on the Flask app or validate `Content-Length` before reading. please check.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,756 @@
...
+        response_time = data.get("response_time_ms", 0)
+        session_id = data.get("session_id")
+
+        if not flashcard_id or is_correct is None:
+            return json_result({"error": "flashcard_id and is_correct are required"}), 400
+
+        user, feature_gate = _current_verbal_flashcards_user()
+        if feature_gate:
+            return feature_gate
+        flashcard = _find_flashcard_for_user(user, flashcard_id)
```
mircealungu: claude says: the flashcard id is `str(bookmark.id)` and `_find_flashcard_for_user` re-runs `get_flashcard_collection(user)` — which calls `BasicSRSchedule.user_words_to_study(user)` — on every submit. If the schedule state changes between the initial GET and submit (e.g. a sibling exercise moves the word out of the 'to study' set), submit returns 404 for a card the user just practiced. Resolve by `user_word_id` + ownership check directly instead of rescanning the session collection. That also fixes the perf concern of re-running the scheduler query once per submit. please check.

asr_service/app.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+        return jsonify(
+            {
+                "success": True,
+                "transcription": transcription,
+            }
+        )
+    except ValueError as exc:
+        return jsonify({"error": str(exc)}), 400
+    except Exception as exc:
```
mircealungu: Status codes between `/health` and `/transcribe` disagree: when the model fails to load, `/health` returns 200 with `'status': 'degraded'`, but `/transcribe` raises `RuntimeError` which hits this generic `except Exception` → 500. Return 503 when `ASR_AVAILABLE` is false (or `asr_model is None`) so orchestrators and load balancers can route around a bad pod.

mircealungu (commented):
Review from a walkthrough of the branch. Most findings are inline below; the main blocker and a few minor items are summarized here.

## 🚨 Blocker — endpoint routes are never registered

`zeeguu/api/endpoints/__init__.py` is the module that imports every endpoint file so the `@api.route(...)` decorators execute. `verbal_flashcards` isn't imported there, so in production **none of `/verbal_flashcards`, `/verbal_flashcards/transcribe`, `/verbal_flashcards/submit`, `/verbal_flashcards/check_pronunciation` exist on the running Flask app.**

The tests pass only because `test_verbal_flashcards.py`'s autouse fixture does `monkeypatch.setattr("zeeguu.api.endpoints.verbal_flashcards...", ...)`, which force-imports the module as a side effect and runs the decorators. In production that side-effect import never happens.

**Fix:** add `from . import verbal_flashcards` to `zeeguu/api/endpoints/__init__.py`.

## Minor items

- `transcribe_audio_endpoint`'s generic `except Exception` echoes `str(e)` back to the client, which can leak internal details (paths, stack-ish info). Log internally and return a generic `"Transcription failed"`.
- `default.env` — the trailing blank-line deletions are unrelated churn; worth isolating or dropping.
- `VERBAL_FLASHCARD_EXERCISE_SOURCE = "Verbal Flashcards"` — check the existing rows in `exercise_source`; other sources may be snake_case and this would create a slightly inconsistent entry.

asr_service/gunicorn.conf.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+
+bind = os.environ.get("GUNICORN_BIND", "0.0.0.0:5002")
+workers = int(os.environ.get("GUNICORN_WORKERS", "1"))
```
mircealungu, line 6: `preload_app = False` + `workers = 1` is fine today, but the model is loaded at module import time in `app.py` (`asr_model = ASRModel.from_pretrained(...)`). With `preload_app = True`, gunicorn imports the app **once in the master process** and then forks workers — on Linux the forked children share the loaded model weights via copy-on-write, so one load serves N workers. With `preload_app = False`, each worker loads its own copy. If you ever bump `workers` above 1 for throughput, memory doubles needlessly. Flipping to `preload_app = True` now makes that a config change, not a code change.

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,117 @@
...
+        mapping[language_code] = service_url
+
+    return mapping
+
+
+def configured_asr_service_urls():
+    """Return configured worker URLs, falling back to the local Danish worker."""
+    raw_value = os.environ.get("ASR_SERVICE_URLS", "")
+
+    if not raw_value and has_app_context():
```
mircealungu: This `has_app_context()` / `current_app.config.get(...)` branch is a third config source that matches neither existing convention in Zeeguu:

- `readability_server` is hardcoded in `zeeguu/core/content_retriever/parse_with_readability_server.py:12`
- `stanza` is env-only (`STANZA_SERVICE_URL` in docker-compose.yml)

Hardcoding doesn't fit here because ASR is language-sharded and the map will grow. The env-var map (`ASR_SERVICE_URLS`) you already use matches the stanza pattern and is the right shape. Suggest dropping this Flask-config branch so env is the single source of truth.

mircealungu (commented):
Two follow-up inline comments: worker model-load semantics, and the Flask-config fallback in the ASR client (context: comparing against readability vs stanza conventions).

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,117 @@
...
+        raw_value = current_app.config.get("ASR_SERVICE_URLS", "")
+
+    if not raw_value:
+        raw_value = LOCAL_DEV_ASR_SERVICE_URLS
+
+    return parse_asr_service_urls(raw_value)
```
mircealungu: Matches the existing convention used by the `stanza` service in this codebase: it reads its URL from a single env var (`STANZA_SERVICE_URL`, set in `docker-compose.yml`) with no Flask-config fallback. The other external service — `readability_server` — hardcodes its URL in `zeeguu/core/content_retriever/parse_with_readability_server.py` and reads no env var at all. Neither of them uses `current_app.config`, so the branch you added here is a third pattern that nothing else in the codebase follows.

Hardcoding (readability's approach) won't work for ASR because the map has to grow per language (`da → asr_da`, `de → asr_de`, …), so env-var-as-map is the right shape. Just drop the Flask-config branch so env is the single source of truth, which mirrors stanza.

Suggested replacement for this function:

```suggestion
def configured_asr_service_urls():
    """Return configured worker URLs, falling back to the local Danish worker."""
    raw_value = os.environ.get("ASR_SERVICE_URLS", "") or LOCAL_DEV_ASR_SERVICE_URLS
    return parse_asr_service_urls(raw_value)
```

Once applied, the `from flask import has_app_context, current_app` line at the top of the file becomes unused — remove it in the same commit.

mircealungu (commented):
Follow-up on #14 — concrete suggestion you can apply directly.

docker-compose.yml
```diff
@@ -139,6 +169,8 @@ services:
       ZEEGUU_DATA_FOLDER: /zeeguu-data/
       ZEEGUU_RESOURCES_FOLDER: /zeeguu-data/
       STANZA_SERVICE_URL: http://stanza:5001
+      ASR_SERVICE_URLS: "${ASR_SERVICE_URLS:-da=http://asr_da:5002}"
```
mircealungu: The current shape `da=http://asr_da:5002` is the worst of both worlds — language is in the host name (`asr_da`) AND in the env-var key (`da=`) AND there's an explicit non-default port. Pick a direction:

- **Option 1: one ASR container, multiple languages.** Container is named `asr`. Language is just a request parameter. URL becomes `da=http://asr` (worker listens on 80 inside the container). Adding German later is just `de=http://asr`.
- **Option 2: one container per language.** Container `asr_da`, with `asr_de` joining later — fine. But drop the explicit port: let each worker listen on 80, so the URL is just `da=http://asr_da`.

Either way, `:5002` is leaking an internal implementation detail into config. Same shape also appears in `default.env:31` — fix both. Not a hard blocker, but worth deciding now while there's only one entry.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,331 @@
...
+def _ensure_verbal_flashcards_enabled(user):
+    if is_feature_enabled_for_user("verbal_flashcards", user):
+        return None
+    return _verbal_flashcards_unavailable_response()
+
+
+def _current_verbal_flashcards_user():
+    user = User.find_by_id(flask.g.user_id)
+    return user, _ensure_verbal_flashcards_enabled(user)
+
```
mircealungu, line 84: `_current_verbal_flashcards_user()` returns `(user, None)` when the user passes the gate and `(user, response_tuple)` when they don't, leading every endpoint to do `if feature_gate: return feature_gate`. The function name promises "current user" but it's actually computing a 404 response on the side. Cleaner to split:

```python
def _current_verbal_flashcards_user():
    return User.find_by_id(flask.g.user_id)
```

…and call `_ensure_verbal_flashcards_enabled(user)` explicitly at the top of each endpoint. Less clever, easier to read.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,331 @@
...
+        if "file" not in request.files:
+            return json_result({"error": "No audio file provided"}), 400
+
+        audio_file = request.files["file"]
+        if audio_file.filename == "":
+            return json_result({"error": "Empty filename"}), 400
+
+        user, feature_gate = _current_verbal_flashcards_user()
+        if feature_gate:
+            return feature_gate
```
mircealungu: The feature gate runs *after* audio-size enforcement, file-presence check, and filename check. Cheap to flip — gate first, then validate. Two reasons:

1. Saves reading a possibly-large audio body for a non-allowlisted user.
2. Leaks "this feature exists" less to outsiders.

Same pattern in `submit_answer` (gate after JSON parse, line 247) and `check_pronunciation` (gate after JSON parse, line 306). Please move all three gates to the top of the endpoint.

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,331 @@
...
+            "flashcards": paginated,
+            "total": total,
+            "limit": limit,
+            "offset": offset,
+        })
+
+    except Exception as e:
+        log(f"Get flashcards error: {e}")
+        traceback.print_exc()
+        return json_result({"error": str(e)}), 500
```
mircealungu: Echoing `str(e)` to the client can leak ORM internals, file paths, and other implementation details. The transcribe endpoint already does this right (returns a static `\"Transcription endpoint error\"` on line 162). Please align the other three the same way: log with detail, respond with a generic message.

Same issue in:
- `submit_answer` (lines 282-284)
- `check_pronunciation` (lines 329-331)

zeeguu/api/endpoints/verbal_flashcards.py
```diff
@@ -0,0 +1,331 @@
...
+
+        flashcard_id = str(data.get("flashcard_id")) if data.get("flashcard_id") is not None else None
+        user_answer = data.get("user_answer", "")
+        is_correct = data.get("is_correct")
+        answer_source = data.get("answer_source", "unknown")
+        response_time = data.get("response_time_ms", 0)
+        session_id = data.get("session_id")
+
+        if not flashcard_id or is_correct is None:
+            return json_result({"error": "flashcard_id and is_correct are required"}), 400
```
mircealungu: Two small things on the `flashcard_id` handling here:

1. `data.get("flashcard_id")` is called twice.
2. The truthiness check `not flashcard_id` would reject `flashcard_id == "0"` (zero is falsy) — never an issue today since IDs are positive, but it's a footgun.

Cleaner:

```python
flashcard_id = data.get("flashcard_id")
if flashcard_id is None or is_correct is None:
    return json_result({"error": "flashcard_id and is_correct are required"}), 400
flashcard_id = str(flashcard_id)
```

zeeguu/core/verbal_flashcards/flashcard_selection.py
```diff
@@ -0,0 +1,111 @@
...
+        "user_word_id": user_word.id,
+        "level": user_word.level,
+        "from": user_word.meaning.origin.content,
+        "to": user_word.meaning.translation.content,
+        "origin": user_word.meaning.origin.content,
+        "translation": user_word.meaning.translation.content,
+        "prompt": prompt,
+        "answer": answer,
+        "expectedText": answer,
+    }
```
mircealungu, line 53: Of the 11 keys returned, the frontend (PR zeeguu/web#1051) reads only **4**: `id`, `prompt`, `answer`, `expectedText`. The other 7 (`bookmark_id`, `user_word_id`, `level`, `from`, `to`, `origin`, `translation`) are never consumed. And `expectedText` and `answer` hold the same value — same field, two names.

Please cut the dict to:

```python
return {
    "id": str(bookmark.id),
    "prompt": prompt,
    "answer": answer,
}
```

…and on the frontend (`VerbalFlashcardsPage.js:967`):

- Replace `card.expectedText` → `card.answer`.
- Delete the `|| currentCard.prompt` fallback. If `answer` is missing, this should fail loudly — the current fallback silently compares the user's spoken word against the *translation* (which is in the user's native language, so the comparison is meaningless).

Heads-up: `submission.py` reads `flashcard["expectedText"]` (line 30) and `flashcard["user_word_id"]` (line 42) internally. When you cut the dict, change `find_flashcard_submission_target` to return the `Bookmark` directly — it already exposes `user_word.meaning.origin.content` (the answer text) and `user_word_id`. That also lets you drop the duplicate ownership check on `submission.py:43-45`, which is already covered inside `find_flashcard_submission_target`.

While you're in there: `find_flashcard_for_user` (lines 68-75 of this file) appears to have no callers anywhere in the repo — safe to delete.

mircealungu:
@jakso535 left a batch of review comments above. Six items, in priority order:

1. **Cut the 11-field card dict to 3** ([F](https://github.com/zeeguu/api/pull/544#discussion_r3163196999)) — touches both repos, see the comment for the coordinated frontend change.
2. **Move feature gate to the top of each endpoint** ([C](https://github.com/zeeguu/api/pull/544#discussion_r3163192939))
3. **Stop returning `str(e)` to clients** ([D](https://github.com/zeeguu/api/pull/544#discussion_r3163193884))
4. **Pick a direction on the ASR container/port shape** ([A](https://github.com/zeeguu/api/pull/544#discussion_r3163190670))
5. **Split `_current_verbal_flashcards_user`** ([B](https://github.com/zeeguu/api/pull/544#discussion_r3163191856))
6. **`flashcard_id` truthiness footgun** ([E](https://github.com/zeeguu/api/pull/544#discussion_r3163195228))

Also: I pushed one commit (`b7cb26f8`) directly to your branch — drops `VERBAL_FLASHCARDS_INVITE_CODES` from `default.env`, `docker-compose.yml`, and the README, since the invite code is now hardcoded in the feature toggle (`spring2026`) and the env var is unused. Pull before continuing.

This is partial review — I've read through endpoint level (file 3 of 7 in the reading plan). More to come on the domain modules, ASR client, and worker.

zeeguu/core/verbal_flashcards/flashcard_selection.py
```diff
@@ -0,0 +1,111 @@
...
+    user_words = BasicSRSchedule.user_words_to_study(user)
+    flashcards = []
+    seen_words = set()
+
+    for user_word in user_words:
+        """
+        Disabled during experimentation, due to uncertainty of participants having level 3 words.
+        """
+        """if (user_word.level or 0) < 3:
+            continue"""
```
mircealungu: Three things stacked here, the first is a live bug:

**1. Level-3 gate is now inconsistent across the module.** You disabled the gate here in `get_flashcard_collection`, but `find_flashcard_submission_target` (line 89) still rejects anything below level 3. Result: under the experiment, a user gets a level-1 or level-2 card in `GET /verbal_flashcards`, practices it, and then `POST /verbal_flashcards/submit` returns 404 ("Flashcard not found"). The experiment is currently broken end-to-end. Either remove the gate from `find_flashcard_submission_target` too, or keep the gate consistent on both sides.

**2. Triple-quoted strings aren't comments.** Lines 48-50 and 51-52 are string literal expressions that evaluate (and are discarded) on every loop iteration. They also won't survive a code formatter that knows to drop unused expressions. Use `#` comments:

```python
# Disabled during experimentation, due to uncertainty of participants
# having level 3 words.
# if (user_word.level or 0) < 3:
#     continue
```

**3. Don't commit "experiment mode" as commented-out code.** This is the kind of thing that silently stays disabled forever. Make it a config knob — e.g., an env var `VERBAL_FLASHCARDS_REQUIRE_LEVEL_3` (default `true`, set to `false` for the experiment). Same applied in both call sites. Then production behavior stays intact, the experiment cohort gets a clean override, and reverting after the experiment is just unsetting the env var instead of going back to find a commented-out block.

zeeguu/api/test/test_verbal_flashcards.py
```diff
@@ -0,0 +1,518 @@
...
+#     expected_prompt = bookmark.user_word.meaning.translation.content
+#     expected_answer = bookmark.user_word.meaning.origin.content
+#     flashcards = client.get("/verbal_flashcards")
+#
+#     assert flashcards["total"] == 1
+#     assert len(flashcards["flashcards"]) == 1
+#     assert flashcards["flashcards"][0]["bookmark_id"] == bookmark_id
+#     assert flashcards["flashcards"][0]["prompt"] == expected_prompt
+#     assert flashcards["flashcards"][0]["answer"] == expected_answer
+#     assert flashcards["flashcards"][0]["expectedText"] == expected_answer
```
mircealungu: This test (and the two further down — `test_transcribe_endpoint_rejects_large_audio_upload` and `test_transcribe_endpoint_returns_503_when_worker_is_not_configured`) was commented out, but each one tests behavior that's still meaningful — the underlying assertions just need updating, not deletion.

**This one (`test_verbal_flashcards_only_returns_level_3_plus_words`):** the new behavior is testable too. Flip the assertion: a level-2 word *should* now appear in the response. That's exactly the contract the experiment relies on, and silently regressing it (e.g., if someone re-enables the gate) is something a test should catch:

```python
def test_verbal_flashcards_returns_words_below_level_3_during_experiment(client):
    _prepare_bookmark_support()
    bookmark_id = add_one_bookmark(client)
    _set_bookmark_level(bookmark_id, 2)

    flashcards = client.get("/verbal_flashcards")
    assert flashcards["total"] == 1
```

**The 413 audio-too-large test:** the status code assertion (`response.status_code == 413`) is still valid and important. Just update the body assertion from `b"Audio upload is too large"` to whatever the new normalized message is — looks like `b"Transcription endpoint rejected large audio upload"`.

**The 503 worker-not-configured test:** same deal. Update the body assertion from `b"No ASR worker configured for language 'de'"` to `b"Transcription endpoint not configured"`. The 503 status check still matters.

Commenting out three tests is throwing away coverage you almost certainly still want. Better to adapt each one in ~30 seconds than to leave them dead in the file.

zeeguu/core/verbal_flashcards/flashcard_selection.py
```diff
@@ -0,0 +1,111 @@
...
+
+        word_text = user_word.meaning.origin.content.lower()
+        if word_text in seen_words:
+            continue
+
+        try:
+            card = _verbal_flashcard_from_user_word(user_word)
+        except Exception as e:
+            log(f"Skipping verbal flashcard for user_word {user_word.id}: {e}")
+            continue
```
mircealungu: `_verbal_flashcard_from_bookmark` already returns `None` for the predictable failure cases (missing bookmark, missing user_word, missing prompt/answer text). What's left for this `except` to catch? Probably ORM lazy-load issues or unicode edge cases — but blanket-catching `Exception` in a tight loop turns real bugs into "skipped card" log lines that nobody monitors.

Either narrow the catch to specific exception types you've actually observed, or drop it — the inner None-return path already handles known bad data. If you want to keep the safety net, at least include a stack trace in the log (`log.exception(...)` or `traceback.format_exc()`) so a real bug doesn't disappear into a one-line warning.

zeeguu/core/verbal_flashcards/text_normalization.py
```diff
@@ -0,0 +1,102 @@
...
+    flashcard scoring was Danish-specific before this registry existed.
+    """
+    if not language_code:
+        return _NORMALIZERS[DEFAULT_NORMALIZER_LANGUAGE]
+
+    normalized_code = str(language_code).casefold()
+    return _NORMALIZERS.get(
+        normalized_code,
+        _NORMALIZERS[DEFAULT_NORMALIZER_LANGUAGE],
+    )
```
mircealungu: Danish should not be the silent fallback for unknown languages. Today, `normalizer_for("de")` returns the Danish normalizer, and the test `test_normalizer_registry_defaults_to_danish_until_more_languages_exist` actively enforces it (`normalizer_for("de").asr_tolerant_form("træ") == "tre"`).

It works today only because no German user reaches this path (they'd 503 at the ASR client first). But the moment a second language ships, this silently applies Danish-specific rules to non-Danish input — strips trailing `g` (`Tag` → `Ta`), folds `ø` → `o`, etc. That's worse than no normalization, and it'll be invisible until someone debugs why German pronunciation scoring is wrong.

Two options:

**Option A — raise on unknown language.** Loud, explicit, makes adding a language a deliberate registry update:
```python
def normalizer_for(language_code):
    code = str(language_code or "").casefold()
    if code not in _NORMALIZERS:
        raise UnsupportedLanguageError(f"No verbal-flashcard normalizer registered for {language_code!r}")
    return _NORMALIZERS[code]
```

**Option B — `GeneralNormalizer` as fallback.** Casefold + Unicode-NFC + whitespace cleanup, no language-specific rewrites:
```python
class GeneralNormalizer:
    def canonical_form(self, word):
        return unicodedata.normalize("NFC", str(word or "").casefold())
    def asr_tolerant_form(self, word):
        return self.canonical_form(word)
    def sanitize_spoken_text(self, text): ...
```

Either way, Danish becomes one *registered* language alongside others, not the implicit default. Update the test (and rename it) to assert the new contract.

zeeguu/core/verbal_flashcards/flashcard_selection.py
```diff
@@ -0,0 +1,111 @@
...
+    if schedule:
+        return schedule
+
+    schedule = FourLevelsPerWord(user_word=user_word)
+    schedule.next_practice_time = datetime.now()
+    schedule.consecutive_correct_answers = 0
+    schedule.cooling_interval = 0
+    db_session.add(schedule)
+    db_session.flush()
+    return schedule
```
mircealungu: What is this function for? It looks like code written to make verbal flashcards target words that have no schedule yet (e.g., during the experiment, when level-1/2 words are surfaced). But manually constructing a schedule and writing arbitrary fields to it (`next_practice_time = now()`, `consecutive_correct_answers = 0`, `cooling_interval = 0`) is bypassing the project's scheduling helpers and risks drifting from the rest of the SR system.

Per the project's CLAUDE.md, the canonical way is `FourLevelsPerWord.find_or_create(db.session, user_word)` — and the comment specifically says **never** hand-roll schedule creation outside of that helper. Two concerns:

1. **Is this experimental scaffolding that snuck into the production code path?** If it exists only to make the experiment work (where you've disabled the level-3 gate), it shouldn't ship as-is. Either gate it behind the same experiment flag (see Comment G), or remove it and rely on the standard scheduler.

2. **If it does need to stay**, switch to `FourLevelsPerWord.find_or_create(...)`. If you have a reason the explicit field assignments here differ from what `find_or_create` does, please document that reason in a code comment — otherwise it'll look like a bug to whoever reads this next.

zeeguu/core/verbal_flashcards/fuzzy_match.py
```diff
@@ -0,0 +1,326 @@
...
+    normalizer = normalizer_for(language_code)
+    normalized_length = len(normalizer.canonical_form(expected_word))
+
+    if normalized_length <= 2:
+        return 1.0
+    if normalized_length == 3:
+        return 0.69
+    if normalized_length == 4:
+        return 0.76
+    return 0.79
```
mircealungu: The values 0.69, 0.76, 0.79 are clearly tuned, but nothing in the file (or in the tests) says where they came from or what false-positive/false-negative rate they target. Numbers like this rot — six months from now it'll be unclear whether tightening or loosening one is safe.

Two requests:

1. Add a comment with the provenance — "calibrated against N samples of Danish ASR output, optimizing for X" or similar.
2. Add at least one regression test per tier: a representative pair that *just* clears the threshold, and one that *just* misses. That way a future change doesn't silently shift acceptance rates without anyone noticing.

zeeguu/core/verbal_flashcards/fuzzy_match.py
```diff
@@ -0,0 +1,326 @@
...
+            normalized_user_word,
+            normalized_expected_word,
+        ),
+        boundary_aware_jaro_winkler_similarity(asr_user_word, asr_expected_word),
+    )
+
+    combined_score = max(
+        normalized_damerau_levenshtein,
+        (normalized_damerau_levenshtein * 0.75) + (jaro_winkler * 0.25),
+    )
```
mircealungu: This formula is hard to reason about. The weighted blend `0.75·DL + 0.25·JW` exceeds `DL` only when `JW > DL`, so `combined_score` is just `DL` whenever `DL ≥ JW`, and the blend otherwise. The whole expression collapses to: "use DL, unless JW happens to be higher, in which case weight them 75/25."

That's an odd rule to motivate. Two cleaner alternatives:

1. **Trust DL alone.** It's already a strong measure; the asymmetric JW boost adds noise more than signal.
2. **Use a fixed weighted average:** `0.75·DL + 0.25·JW` for all cases. This treats JW as a consistent secondary signal.

Either is more interpretable than `max(DL, blend)`. If there's a reason for the "JW only counts when it pulls up" behavior, please document it inline — otherwise the next reader (likely future-you) will assume it's a bug.

zeeguu/core/verbal_flashcards/fuzzy_match.py
```diff
@@ -0,0 +1,326 @@
...
+        "feedback": feedback,
+        "wordMatches": word_matches,
+    }
+
+
+def get_feedback_message(accepted_words, total_words):
+    """Return one of the two simplified feedback outcomes for verbal flashcards."""
+    if total_words and accepted_words == total_words:
+        return "Success"
+    return "Very close, try again"
```
mircealungu, line 431: A learner who got 0 of 3 words correct currently sees "Very close, try again." That's the wrong nudge — they weren't close, and pretending otherwise makes the feedback dishonest.

Add at least a "didn't catch that" tier:

```python
def get_feedback_message(accepted_words, total_words):
    if not total_words:
        return ""
    if accepted_words == total_words:
        return "Success"
    if accepted_words == 0:
        return "Didn't catch that, try again"
    return "Very close, try again"
```

Three tiers covers the realistic outcome buckets without becoming chatty.

zeeguu/core/verbal_flashcards/fuzzy_match.py
```diff
@@ -0,0 +1,326 @@
...
+
+    return {
+        "accuracy": word_accuracy,
+        "wordAccuracy": word_accuracy,
+        "acceptedAccuracy": accepted_accuracy,
+        "acceptedWordCount": accepted_words,
+        "isAccepted": is_accepted,
+        "feedback": feedback,
+        "wordMatches": word_matches,
+    }
```
mircealungu, line 372: `accuracy` and `wordAccuracy` are the same value (`word_accuracy`). Looks like a copy-paste artifact from when the response shape was being figured out.

Pick one. Earlier audit of the frontend (PR zeeguu/web#1051) showed it reads `accuracy` for the status bar but not `wordAccuracy` — drop the unused alias and keep the API surface tight.

zeeguu/core/verbal_flashcards/fuzzy_match.py
```diff
@@ -0,0 +1,326 @@
...
+    normalizer_for,
+    sanitize_spoken_text,
+)
+
+
+FUZZY_ACCEPTANCE_BUFFER = 0.1
+
+
+def damerau_levenshtein_distance(source, target):
+    """Classic dynamic-programming Damerau-Levenshtein distance."""
```
mircealungu: The function name and docstring both say "Damerau-Levenshtein", but the implementation is actually **OSA (Optimal String Alignment)** distance. The difference: OSA forbids editing the same substring more than once; true Damerau-Levenshtein allows it.

For short flashcard words the practical difference is invisible — but the docstring should match what the code does. Either:

- Rename the function to `optimal_string_alignment_distance` and update the docstring (correct, but ripples through call sites), or
- Just adjust the docstring to "OSA / restricted edit distance" and note it's the textbook implementation people usually mean by "Damerau-Levenshtein."

Pedantic, but the kind of thing that wastes a future debugger's hour when they expect true DL behavior and don't get it.

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,119 @@
+import os
+import logging
+
+import requests
+
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ASR_SERVICE_TIMEOUT = float(os.environ.get("ASR_SERVICE_TIMEOUT", "30"))
+LOCAL_DEV_ASR_SERVICE_URLS = "da=http://127.0.0.1:5002"
```
mircealungu: Quick coordination note — this dev fallback couples to two earlier comments:

- **Comment A** (docker-compose port shape): if the prod URL drops the explicit `:5002` and the worker listens on 80, this becomes `da=http://127.0.0.1`.
- **Comment J** (Danish-as-fallback footgun): hardcoding Danish at the dev layer mirrors the same blind spot from `text_normalization.py`. Whatever the production registry decision is (raise on unknown / `GeneralNormalizer` fallback), the dev defaults should follow it.

Not a separate issue — just "don't let dev defaults drift away from the production decisions" when fixing A and J.

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+        mapping[language_code] = service_url
+
+    return mapping
+
+
+def in_development_context():
+    """Return True when local dev defaults should be enabled."""
+    flask_env = os.environ.get("FLASK_ENV", "").strip().casefold()
+    return flask_env == "development"
```
mircealungu: `FLASK_ENV` was removed in Flask 2.3 (April 2023). The project's `requirements.txt` pins `Flask>=3.0.0`, so this name is no longer recognized by Flask itself — and grep across `zeeguu/` shows this is the **only** file in the repo that reads `FLASK_ENV`. Zero precedent in the codebase, deprecated upstream — so it's not following a local convention or a Flask one.

Two cleaner options:

1. Use `FLASK_DEBUG` (still supported in Flask 3): `os.environ.get("FLASK_DEBUG") == "1"`.
2. Use a Zeeguu-specific name like `ZEEGUU_ENV=development`, which makes the intent unambiguous and avoids the dead Flask convention entirely.

Either way is better than the current name, which will mislead anyone who reads it into thinking Flask is doing something with the variable.

zeeguu/core/audio_lessons/asr_service_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return parse_asr_service_urls(raw_value)
+
+
+def get_asr_service_url(language_code, service_url_map=None):
+    if not language_code:
+        return None
+
+    service_url_map = service_url_map or configured_asr_service_urls()
+    return service_url_map.get(str(language_code).casefold())
```
mircealungu: This re-reads `os.environ` and re-runs `parse_asr_service_urls` on every transcribe call. Two reasons to fix:

1. Tiny perf — useless work on a hot endpoint.
2. Behavior depends on whether someone mutated `os.environ` mid-process, which can happen in tests but shouldn't matter in prod.

Parse once at module load, the same way `DEFAULT_ASR_SERVICE_TIMEOUT` is captured at import:

```python
ASR_SERVICE_URL_MAP = parse_asr_service_urls(
    os.environ.get("ASR_SERVICE_URLS", "")
    or (LOCAL_DEV_ASR_SERVICE_URLS if in_development_context() else "")
)

def get_asr_service_url(language_code, service_url_map=None):
    if not language_code:
        return None
    return (service_url_map or ASR_SERVICE_URL_MAP).get(str(language_code).casefold())
```

The `service_url_map` parameter override stays so tests can inject their own map.

asr_service/app.py
```diff
@@ -0,0 +1,176 @@
...
+@app.route("/health", methods=["GET"])
+def health():
+    return jsonify(
+        {
+            "status": "ok" if ASR_AVAILABLE and asr_model is not None else "degraded",
+            "worker_name": ASR_WORKER_NAME,
+            "worker_language": ASR_LANGUAGE_CODE,
+            "model_loaded": asr_model is not None,
+        }
+    )
```
mircealungu: The HTTP status is always 200, even when `ASR_AVAILABLE` is False (model failed to load). The Dockerfile's healthcheck is:

```
HEALTHCHECK ... CMD curl -f http://localhost:5002/health || exit 1
```

`curl -f` only fails on 4xx/5xx — so a degraded worker reports as **healthy**. Container orchestration (Docker healthcheck, Kubernetes liveness probes) will leave a permanently-broken worker running, silently returning 503 to every transcribe call.

Real bug. Fix:

```python
@app.route("/health", methods=["GET"])
def health():
    payload = {
        "status": "ok" if ASR_AVAILABLE and asr_model is not None else "degraded",
        "worker_name": ASR_WORKER_NAME,
        "worker_language": ASR_LANGUAGE_CODE,
        "model_loaded": asr_model is not None,
    }
    status_code = 200 if payload["status"] == "ok" else 503
    return jsonify(payload), status_code
```

asr_service/requirements.txt
```diff
@@ -0,0 +1,5 @@
+flask>=2.0.0
+gunicorn>=21.0.0
+huggingface_hub>=0.24.0
+nemo_toolkit[asr]
```
mircealungu: `nemo_toolkit[asr]` is unpinned. NeMo is a fast-moving framework with frequent breaking changes around model output shapes — the transcript parsing in `app.py:95-113` is already shotgun-defensive precisely because the output shape isn't stable across versions (handles tuples, raw strings, lists of objects with `.text`, lists of lists).

The moment NeMo ships a new release that introduces a fifth shape, this worker silently degrades — caught by the bare `except`, returned as 500. A build six months from now would pull a different NeMo than today's, with no signal that anything changed.

Pin it. Either an exact version (`nemo_toolkit[asr]==2.0.0`) or a tight range (`nemo_toolkit[asr]>=2.0,<2.1`). Same applies to `pydub` and `huggingface_hub` — both unpinned today, and `huggingface_hub` is what fetches the actual model weights, so reproducibility depends on it too.

Pin everything you don't want surprising you.

asr_service/app.py
```diff
@@ -0,0 +1,176 @@
...
+            }
+        )
+    except ASRAudioTooLarge as exc:
+        return jsonify({"error": str(exc)}), 413
+    except ASRModelUnavailable as exc:
+        return jsonify({"error": str(exc)}), 503
+    except ValueError as exc:
+        return jsonify({"error": str(exc)}), 400
+    except Exception as exc:
+        return jsonify({"error": str(exc)}), 500
```
mircealungu: Same `str(exc)` leak pattern as [Comment D](https://github.com/zeeguu/api/pull/544#discussion_r3163193884) on the main API. Although this worker is only called by your own backend, the worker's exception string propagates through `ASRServiceRequestError` and lands in the user's HTTP response body if the main API still has the leak from Comment D.

Fix the same way: log with detail, respond with a generic message like `"Transcription failed"`.

asr_service/app.py
```diff
@@ -0,0 +1,176 @@
...
+            nested = first[0]
+            if hasattr(nested, "text"):
+                transcription = nested.text
+            elif isinstance(nested, str):
+                transcription = nested
+
+        if transcription is None:
+            raise TypeError(
+                f"Unexpected transcription output: {type(transcript)} / {type(first)}"
+            )
```
mircealungu: This is shotgun parsing — four different output shapes handled, clearly evolved through experimentation across NeMo versions. Two issues:

1. Without comments explaining which shape comes from which model/version, the next reader has no way to maintain or extend it.
2. The first NeMo release that introduces a fifth shape silently breaks this (caught by the bare `except`, returned as 500). The user just sees "transcription failed" with no idea why.

Two ways to clean this up:

- **Document inline** which shape each branch handles: `# tuple of (transcripts, log_probs) — older NeMo API`, `# Hypothesis object — current API`, etc. At least future-you knows what each branch is for.
- **Target one shape and reject the rest.** Combined with [Comment V](https://github.com/zeeguu/api/pull/544#discussion_r3163573943) (pinning NeMo), commit to the one output shape your pinned version produces. Anything else becomes an upgrade-time problem, not a production-time problem.

Long-term, the second is the right move.

asr_service/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY asr_service/requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY asr_service/app.py asr_service/gunicorn.conf.py ./
+
+EXPOSE 5002
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=300s --retries=3 \
+    CMD curl -f http://localhost:5002/health || exit 1
```
mircealungu: The port `5002` is hardcoded in four places:

```
asr_service/app.py:176        port=int(os.environ.get("ASR_SERVICE_PORT", "5002"))
asr_service/Dockerfile:17     EXPOSE 5002
asr_service/Dockerfile:20     HEALTHCHECK ... curl -f http://localhost:5002/health
asr_service/gunicorn.conf.py:4  bind = "0.0.0.0:5002"
```

Couples to [Comment A](https://github.com/zeeguu/api/pull/544#discussion_r3163190670). Whatever the decision is (drop to 80, keep 5002, parameterize), defining it once and threading it through is the right shape — not duplicating the literal four times.

Cleanest: a single `ARG ASR_SERVICE_PORT=5002` (or whatever) at the top of the Dockerfile, used in `ENV`, `EXPOSE`, and the HEALTHCHECK URL. The gunicorn config and `app.py` already read `ASR_SERVICE_PORT` env var (well, `app.py` does — `gunicorn.conf.py` reads `GUNICORN_BIND` separately, which should also be unified).

If the realistic path is one-container-per-language (each container running on the same port, container name as routing key), then this is even more important: when `asr_en` joins, you don't want to track down four files to make sure it lines up with `asr_da`.

mircealungu:
## Forward-compatibility for a likely Whisper switch

The per-language container architecture works for today's Danish-only experiment, but it's worth noting that scaling beyond Danish probably *won't* mean adding more Parakeet workers — there's no `parakeet-rnnt-110m` for the other 14 Zeeguu languages. The realistic next step is a single multilingual model (Whisper-large or Faster-Whisper-large) covering all languages in one container.

**The good news: most of this PR's design layers are forward-compatible.**

- ✅ Main API treats ASR as a black box — only knows "language_code in, transcription out".
- ✅ Text normalization is a per-language registry, not hardcoded into callers.
- ✅ Fuzzy matching is ASR-agnostic.
- ✅ Domain layer (`flashcard_selection`, `submission`) doesn't know ASR exists.

The contract `POST /transcribe { audio, language_code } → { transcription }` is exactly right — Whisper, Parakeet, AWS Transcribe all fit it. Don't change the contract.

**What does leak Parakeet/per-language assumptions and is worth tightening:**

**1. `ASR_SERVICE_URLS` shape (`da=URL,de=URL,fr=URL,...`)** is the strongest leak. The format mandates per-language URL configuration even when there's only one backend. With Whisper you'd have to configure 15 entries pointing to the same URL, which is awkward and signals "we built around the wrong abstraction." Forward-compatible shape:

```
ASR_SERVICE_URL=http://asr                  # common case — one backend, all languages
ASR_LANGUAGE_OVERRIDES=da=http://asr-da     # only when you really need a per-language override
```

Default to single URL; per-language overrides are the exception.

**2. Worker's `ASR_LANGUAGE_CODE` + rejection at `app.py:73-77`** encodes "this worker handles exactly one language." Trivial loosening: `ASR_SUPPORTED_LANGUAGES` (comma-separated, defaults to one). The rejection logic becomes `requested_language_code not in SUPPORTED`. No operational difference today, no lock-in tomorrow.

**3. README in `asr_service/`** documenting the contract: "this is one possible ASR backend; the contract is POST /transcribe with audio + language_code, returning {transcription}". Future-you (or whoever swaps in Whisper) reads that and knows what surface to preserve.

None of these are PR-blockers. The current design ships fine for the Danish experiment. They're the kind of thing where doing them now costs ~30 minutes and doing them later (after another language is added with the wrong shape) costs hours of cleanup.

jakso535:
# Verbal Flashcards Change Summary

## Flashcard Response Shape

- Reduced each verbal flashcard payload to the fields used by the frontend:
  - `id`
  - `prompt`
  - `answer`
- Removed duplicated/unused fields from the response shape, including:
  - `expectedText`
  - `bookmark_id`
  - `user_word_id`
  - `level`
  - `from`
  - `to`
  - `origin`
  - `translation`
- Updated the frontend verbal flashcards page to read `card.answer` instead of
  `card.expectedText`.
- Removed the frontend fallback that compared spoken text against the prompt
  when the answer was missing.
- Changed submission lookup so the backend resolves the submitted flashcard by
  bookmark directly instead of depending on the public flashcard dictionary.
- Removed an unused flashcard lookup helper after confirming it had no callers.

## Endpoint Safety And Error Handling

- Moved verbal flashcard feature-gate checks to the top of the relevant
  endpoints before parsing bodies or validating uploaded audio.
- Stopped returning raw exception text to clients in verbal flashcard endpoint
  errors.
- Replaced public `str(exc)` responses with generic, stable error messages.
- Kept detailed error information in server-side logs where useful.
- Updated `submit_answer` validation so `flashcard_id` is read once, checked
  explicitly against `None`, and then coerced to string.
- Kept `session_id` validation explicit while avoiding exception echoing.

## Flashcard Selection And Submission Flow

- Made the level-3 word requirement configurable with
  `VERBAL_FLASHCARDS_REQUIRE_LEVEL_3`.
- Restored production default behavior to require level-3 words.
- Allowed experiments to include lower-level words by setting the env var to
  false.
- Made the level gate consistent between flashcard selection and submission.
- Replaced commented-out experiment code with the config flag.
- Switched schedule creation to the canonical scheduler helper:
  `FourLevelsPerWord.find_or_create(db.session, user_word)`.
- Removed hand-rolled schedule field initialization.
- Changed predictable bad flashcard data handling to return `None`, while
  avoiding blanket exception swallowing in the selection loop.

## Text Normalization

- Updated the per-language normalizer registry.
- Made unknown normalizer languages fail loudly instead of silently using Danish
  rules.
- Updated tests to assert that unknown languages raise an unsupported-language
  error.
- Kept Danish-specific normalization isolated behind the registry.

## Fuzzy Matching And Pronunciation Scoring

- Reworked scoring to use Optimal String Alignment edit distance as the
  acceptance criterion.
- Renamed/documented the previous Damerau-Levenshtein implementation as OSA /
  restricted edit distance.
- Kept Jaro-Winkler as a diagnostic signal rather than part of the acceptance
  decision.
- Simplified the acceptance policy:
  - words of length 1-2 require exact match
  - words of length 3+ allow one OSA edit
- Added/updated regression tests for accepted and rejected edit-distance cases.
- Removed the duplicate `wordAccuracy` response field and kept `accuracy`.
- Added a three-tier feedback message policy:
  - all words accepted: success-style feedback
  - no words accepted: "Didn't catch that, try again"
  - partial match: "Very close, try again"

## ASR Client Configuration

- Removed Flask app-config fallback from the ASR service client.
- Kept ASR configuration as environment-driven, matching the existing service
  convention used elsewhere in the codebase.
- Replaced deprecated `FLASK_ENV` local-dev detection with `FLASK_DEBUG`.
- Parsed ASR configuration at module import instead of on every transcription
  call.
- Changed ASR routing config from language-map-first:

  ```env
  ASR_SERVICE_URLS=da=http://asr
  ```

  to the forward-compatible shape:

  ```env
  ASR_SERVICE_URL=http://asr
  ASR_LANGUAGE_OVERRIDES=
  ```

- Kept per-language overrides available only for exceptions, for example:

  ```env
  ASR_LANGUAGE_OVERRIDES=da=http://asr-da
  ```

- Added a local debug fallback to `http://127.0.0.1:5002` only when
  `FLASK_DEBUG=1`.
- Updated `start.py` to set `FLASK_DEBUG=1` for local `python start.py` runs so
  the local ASR worker fallback works.

## ASR Worker Architecture

- Kept the ASR worker contract generic:
  - request: audio file plus `language_code`
  - response: `{ "success": true, "transcription": "..." }`
- Added `ASR_SUPPORTED_LANGUAGES`, allowing:
  - one language such as `da`
  - a comma-separated list
  - `*` for a future multilingual worker
- Changed the worker health endpoint to return HTTP `503` when the model is not
  loaded.
- Updated health payload from singular `worker_language` to
  `worker_languages`.
- Kept Docker production routing on the service name `asr` without leaking
  `:5002` into production config.
- Kept direct local worker execution available on port `5002`.
- Unified Docker worker port handling around `ASR_SERVICE_PORT`.
- Kept gunicorn bound to `ASR_SERVICE_PORT`.
- Added ASR service documentation describing the worker contract and config
  model.

## ASR Worker Dependencies And NeMo Output

- Pinned ASR worker dependencies to the versions currently verified locally:
  - `nemo_toolkit[asr]==2.7.3`
  - `huggingface_hub==0.36.2`
  - `pydub==0.25.1`
- Verified the pinned NeMo Danish Parakeet model returns `list[Hypothesis]`.
- Replaced shotgun transcription output parsing with an explicit parser for the
  pinned output shape.
- Made unexpected NeMo output shapes fail loudly instead of being guessed at.
- Added ASR worker tests for the accepted Hypothesis shape and rejected legacy
  shapes.

## Docker And Environment

- Updated `docker-compose.yml` so the main API uses:

  ```env
  ASR_SERVICE_URL=http://asr
  ASR_LANGUAGE_OVERRIDES=
  ```

- Updated `default.env` to the same config shape.
- Kept the ASR service container named `asr`.
- Kept internal container port usage at `80` for production-style Docker
  routing.
- Added `ASR_SUPPORTED_LANGUAGES=da` to the current ASR worker environment.
- Confirmed `docker compose config` is valid after the config changes.

## Tests And Verification

- Restored previously commented-out tests by updating their assertions instead
  of leaving coverage disabled.
- Added tests for:
  - lower-level experiment flashcards
  - ASR config parsing and routing
  - local debug ASR fallback
  - ASR worker health status codes
  - ASR worker supported-language checks
  - ASR worker generic error responses
  - NeMo output shape extraction
  - OSA scoring behavior
- Repeatedly verified the focused suites during the work:
  - `asr_service/test_app.py`
  - `zeeguu/api/test/test_verbal_flashcards.py`

## Notes

- The local direct ASR worker port `5002` remains only for development.
- Production ASR routing should use `ASR_SERVICE_URL` and optional
  `ASR_LANGUAGE_OVERRIDES`.

jakso535:
Verbal flashcards now use Meaning pairs as possible answer also. This can in some cases fix translations errors of:

"Ball" being translated to "bolden" instead of "bold", where the bookmark variant contains "bold".

zeeguu/core/verbal_flashcards/flashcard_selection.py
```diff
@@ -0,0 +1,142 @@
...
+    return _verbal_flashcard_from_bookmark(user_word.preferred_bookmark)
+
+
+def _add_unique_text(texts, text):
+    cleaned_text = str(text or "").strip()
+    if cleaned_text and cleaned_text.casefold() not in {t.casefold() for t in texts}:
+        texts.append(cleaned_text)
+
+
+def answer_variants_for_bookmark(bookmark):
```
mircealungu, line 90: Nice idea for the inflection case — `bold` / `bolden` sharing the cue "ball" is exactly where strict matching frustrates learners.

But the current rule (any non-INVALID `Meaning` with same origin-language, translation-language, and lowercased translation text) can't distinguish inflectional variants from genuine homonyms. Concrete Danish example, for an English speaker:

- `spring → forår` (the season)
- `spring → fjeder` (a mechanical spring)
- `spring → kilde` (a water source)

All three would now be accepted as correct answers for the cue "spring", regardless of which sense the learner actually scheduled. If they specifically picked `spring → fjeder` to drill engineering vocabulary, getting credit for `kilde` undermines the study.

To do this reliably, I think we need a **meaning-family / inflection-group** relationship in the data model — something that marks `bold` and `bolden` as forms of the same lemma, while keeping `fjeder` and `kilde` as unrelated meanings that just happen to share an English cue. Matching on cue text alone over-accepts on homonyms.

Until that data exists, one safer scoping option: limit variants to the same `user_word.meaning.origin.lemma` (if available), or only accept variants when the edit distance between origin contents is small (which would catch `bold` ↔ `bolden` but reject `fjeder` ↔ `kilde`). Worth thinking about before this lands as the default scoring policy.

jakso535: Good catch! This should be fixed now. The intermediary decision is to accept variants only if they are 2 or fewer edits from the original bookmark :)

mircealungu:
**Small follow-up on the ASR Dockerfile build context** (was going to leave this as a line comment on `asr_service/Dockerfile` but the API rejected it):

The Dockerfile uses `asr_service/`-prefixed COPY paths, which forces the whole api repo to be the docker build context. Aligning with how `stanza_service` is built (its own directory as the build context) keeps the worker self-contained.

```diff
-COPY asr_service/requirements.txt .
+COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt

-COPY asr_service/app.py asr_service/gunicorn.conf.py ./
+COPY app.py gunicorn.conf.py ./
```

And `docker-compose.yml` simplifies to:

```diff
   asr:
-    build:
-      context: .
-      dockerfile: ./asr_service/Dockerfile
+    build: ./asr_service
```

Smaller build context (faster builds, no risk of bundling unrelated api code into the image layer), and the deployment compose in `ops/` can mirror the stanza pattern: `build: ${ZEEGUU_SOURCE_FOLDER}/api/asr_service`.

PR #564, opened 2026-04-23
Funnier badge descriptions for first three levels

mircealungu:
## Summary
- Rewrites the `description` on levels 1–3 of every badge category in `26-02-19-b--insert_default_badges.sql` so each badge reads as specific copy tied to its name and threshold (e.g. *"Five hundred words decoded. The dictionary is starting to sweat."*), instead of the generic `Translate {threshold} unique words while reading.` template that repeats across all five levels.
- Levels 4–5 are left on the templated placeholder for now — scope is only the first three per category so the team can sanity-check the voice before we do the upper tiers.
- Addresses the feedback on PR #517 about the `{threshold}`-style descriptions looking unprofessional in the UI (e.g. "Add 1 friend(s).", "Complete 1 audio lesson(s).").

## Test plan
- [ ] Eyeball the copy per-category and flag any tone/voice changes
- [ ] Decide whether to extend the same treatment to levels 4 and 5 now or in a follow-up
- [ ] Run the migration against a throwaway DB and render the badges in the UI to confirm no quoting/escaping surprises with the apostrophes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #573, opened 2026-04-25
Added default descriptions to badge_category

gabortodor:
Added singular and plural default descriptions to badge categories which can be used as badge descriptions on the frontend for not yet achieved badges.

Also added funny descriptions for badges, which will be shown once you achieve them. Took the names from https://github.com/zeeguu/api/pull/564, and added names for level 4 and 5 as well.

Also, for friend users, only achieved badges are returned by the api from now on.

### DIFF
--- a/tools/migrations/26-04-25--add_default_description_to_badge.sql
+++ b/tools/migrations/26-04-25--add_default_description_to_badge.sql
@@ -1,42 +1,67 @@
--- Add default descriptions for badge categories (singular + plural forms)
--- These are used as fallback templates when a specific badge level
--- is not yet achieved. They support {threshold} substitution for progress display.
+-- Add default descriptions directly on badges.
+-- These are used as fallback descriptions before a badge is achieved.
+-- Each description already includes the correct threshold value.
 
-ALTER TABLE badge_category ADD COLUMN singular_default_description TEXT DEFAULT NULL;
-ALTER TABLE badge_category ADD COLUMN plural_default_description TEXT DEFAULT NULL;
+ALTER TABLE badge ADD COLUMN default_description TEXT;
 
-UPDATE badge_category
-SET
-  singular_default_description = CASE id
-    WHEN 1 THEN 'Translate a unique word while reading.'
-    WHEN 2 THEN 'Solve an exercise correctly.'
-    WHEN 3 THEN 'Complete an audio lesson.'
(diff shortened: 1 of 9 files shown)


mircealungu:
Suggestion: put `default_description` on `badge` instead of splitting it into `singular_default_description` + `plural_default_description` on `badge_category`.

Each badge row already knows its own threshold and level, so:

- No `{threshold}` placeholder — write the literal string ("Translate 10 unique words while reading.").
- No singular/plural column split — pick the right phrasing per row ("Learn a new word." vs "Learn 10 new words.").
- No unreachable strings (currently 4/7 singular descriptions can never render because the lowest threshold in those categories is >1).
- Symmetric API: every badge gets `description` (achieved copy) and `default_description` (unachieved copy). Client logic becomes `achieved ? description : default_description` — no template substitution, no threshold inspection, no singular/plural rule spread across web/iOS/Android.

Migration shrinks to one `ALTER TABLE badge ADD COLUMN default_description TEXT NOT NULL` + one `UPDATE badge SET default_description = CASE id WHEN ... END`. `BadgeCategory` model and serializer stay untouched.

mircealungu:
One more naming nit before this lands: the pair `description` / `default_description` reads ambiguously.

Suggest renaming both to make the state explicit:

- `default_description` → `unachieved_description`
- `description` → `achieved_description`

That way the columns describe *when* each is shown, and there's no chance of a future reader picking the wrong one.

Cheap to do while the PR is still open — rename in the migration, the `Badge` model, the serializer key, and the test fixture. And I guess in the frontend too.

This code here (https://github.com/zeeguu/web/pull/1084/changes) `description: lvl.achieved ? lvl.description : lvl.default_description,` will read nicer after the refactoring.

klnyzzz33:
This should be fixed now and ready to merge @mircealungu

klnyzzz33:
I also added the implementation of the unachieved badges not being displayed for friends (mentioned in https://github.com/zeeguu/web/pull/1085) to this PR, because I had to solve merge conflicts and it was easier this way.

mircealungu:
added https://github.com/zeeguu/api/pull/573/changes/d04021e1fa83c3ae0807416044bb38b469a2f1ee to maintain backwards compatibility with old apps that take a while until they're migrated

PR #582, opened 2026-04-28
Add anonymous account to cohort on account upgrade.

xXPinkmagicXx:
**Problem**
On mobile when you use an invite code it is not saved on the user account and the user is not added to the cohort/class room.
This is a bad user experience as you have to add the invite code twice.

**Solution**
If you are using an invite code on mobile it is saved on the user.
When you upgrade from the anonymous user to a real user the invite code from the user is used to add the real user to the cohort /classroom. This is the desired behavior.

Note: It does not make sense to add a anonymous user to a classroom and therefore we wait until account upgrade.


mircealungu:
nice thinking!

PR #589, opened 2026-04-29
Removed uuid display name from anonymous account creation

gabortodor:
Now the anonymous user creation flow does not set the display name to an uuid


mircealungu:
tak @gabortodor

Issue #592, opened 2026-04-30
Email Verification fails when email does not match casing.


xXPinkmagicXx:
When upgrading your account if you write Abc@email.com instead of abc@email.com 
The email varification will fail.

```py
    @classmethod
    def find_last_code(cls, email):
        """Return the last code object for the given email, or None if not found"""
        return cls.query.filter(cls.email == email).order_by(cls.time.desc()).first()
```
You get this error but only when you are trying to confirm you password. This is not intuitive. As the correct error messages is not displayed either
<img width="774" height="422" alt="Image" src="https://github.com/user-attachments/assets/47dd6f94-c03c-42c6-9938-7143170f8316" />

xXPinkmagicXx:
@mircealungu see fix in #599

PR #599, opened 2026-05-06
Fix: issue 592 case sensitive email (with migration)

xXPinkmagicXx:
## Problem
On the frontend the email verification fails in the upgrade account modal if the email is not lower cased.
#592 

## Testing
- [x] Test frontend locally


mircealungu:
merged and run the migration @xXPinkmagicXx .

PR #601, opened 2026-05-07
Verbal flashcards word length dependent permission

jakso535:
Permission of what is correct or incorrect now depends on the length of the word. That way we are still pretty strict about short words, but for words like "anderledes" we can now accept variants like "anderlas" "anerles" which makes the feature a bit more forgiving.


mircealungu:
Nice. This will probably improve a bit the accuracy!

PR #603, opened 2026-05-12
name change for set_onboarding_message_click_time

AnnieeBennie:
changed the name for set_onboarding_message_click_time to mark_onboarding_message_dismissed

### DIFF
--- a/tools/migrations/26-05-03--add_user_onboarding_message_table.sql
+++ b/tools/migrations/26-05-03--add_user_onboarding_message_table.sql
@@ -3,7 +3,7 @@ CREATE TABLE `zeeguu_test`.`user_onboarding_message` (
     `user_id` INT NOT NULL,
     `onboarding_message_id` INT NOT NULL,
     `message_shown_time` DATETIME NULL,
-    `message_dismissed_time` DATETIME NULL,
+    `message_click_time` DATETIME NULL,
     PRIMARY KEY (`id`),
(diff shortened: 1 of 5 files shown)


mircealungu:
@AnnieeBennie heads up — I just pushed a small commit (`c4a91be3`) to this branch to add a forward migration.

Background: the existing `26-05-03--add_user_onboarding_message_table.sql` is already on `master` and has run on production. When a migration has already executed, editing the file in place (which is what this PR does) doesn't re-apply on databases that already ran it — the migration runner sees it as "done" and skips. 

So we need an additional migration, that will ALTER the table to do the rename. 

The new file `tools/migrations/26-05-14--rename_message_click_time_to_dismissed.sql` does an `ALTER TABLE … CHANGE COLUMN` so existing environments converge. Fresh environments still get the renamed column directly from the edited `26-05-03` file, so both paths arrive at the same end state.

General rule of thumb: once a migration is on `master` (and likely deployed), treat it as immutable history. Any subsequent schema change needs a new dated migration.

Also worth doing in this PR (optional, small):
- Rename the helper `update_user_onboarding_message_time` → `set_message_dismissed_time` to mirror `set_message_shown_time`.

mircealungu:
Actually, please @AnnieeBennie do this:
- Rename the helper update_user_onboarding_message_time → set_message_dismissed_time to mirror set_message_shown_time.

AnnieeBennie:
@mircealungu I updated the name

PR #649, opened 2026-06-05
Strip tracking params (gaa_*, utm_*, ...) from stored URLs

mircealungu:
## Problem

The bt.dk URL below broke article creation (reported in prod logs):

```
https://www.bt.dk/krimi/rumaenske-tyve-haerger--nu-doemmes-fire-om-dagen?gaa_at=...&gaa_n=...&gaa_ts=...&gaa_sig=...
```

Some publishers (e.g. bt.dk via Google Discover / "Subscribe with Google") append long signed **Google Article Access** tokens — `gaa_at`, `gaa_n`, `gaa_ts`, `gaa_sig` — to article URLs. The full URL is 278+ chars, and that overflowed two 255-char columns:

1. **`user_activity_data.value`** — the `OPEN POPUP` event from the extension 500'd with `DataError 1406: Data too long for column 'value'`, killing the whole `/upload_user_activity_data` request.
2. **`url.path`** (via `Url.get_path`) — the over-long path silently failed the `len(path) > 255` guard, so the article's URL row was dropped.

## Fix

New helper `remove_tracking_query_params()` (`zeeguu/core/util/url.py`) strips known tracking cruft (`gaa_*`, `utm_*`, `fbclid`, `gclid`, `_ga`), keeps real params, and leaves non-URL strings untouched. Applied at both sites:

- `create_from_post_data`: clean + clamp `value` to 255 (backstop against any future monster URL).
- `Url.get_path`: clean at the single choke point used by `__init__`, `find_or_create`, and `find`, so store and lookup stay canonical and consistent.

The bt.dk URL drops from 278 → 72 chars (clean canonical article URL).

## Tests

- `test_util_url.py` — strips gaa tokens, strips utm/click-ids while keeping real params, leaves clean/non-URL/empty strings untouched.
- `test_url.py` — `get_path` strips tracking params.

All pass. Verified the full model package still imports cleanly (no circular-import regression).

### DIFF
--- a/zeeguu/core/model/url.py
+++ b/zeeguu/core/model/url.py
@@ -13,6 +13,7 @@ import random
 from zeeguu.core.model.db import db
 
 from zeeguu.core.model.domain_name import DomainName
+from zeeguu.core.util.url import remove_tracking_query_params
 
 
 class Url(db.Model):
@@ -78,6 +79,11 @@ class Url(db.Model):
 
     @classmethod
     def get_path(cls, url: str):
+        # Strip tracking cruft (gaa_*, utm_*, ...) before extracting the path,
+        # so the stored path stays canonical and short. Some publishers append
+        # long signed access tokens that otherwise overflow the 255-char column.
+        url = remove_tracking_query_params(url)
+
         protocol_re = "(.*://)?"
         domain_re = "([^/?]*)"
         path_re = "(.*)"
--- a/zeeguu/core/model/user_activitiy_data.py
+++ b/zeeguu/core/model/user_activitiy_data.py
@@ -24,6 +24,7 @@ from zeeguu.core.constants import (
 from zeeguu.core.behavioral_modeling import (
     find_last_reading_percentage,
 )
+from zeeguu.core.util.url import remove_tracking_query_params
 import zeeguu
 
 from zeeguu.core.model.db import db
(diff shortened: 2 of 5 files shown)


mircealungu:
### Follow-up after self code-review

Ran `/code-review` (high effort) on this branch. It flagged that the first cut used a `parse_qsl`→`urlencode` round-trip that **mutated URLs even when no tracking param was present**:

- `articleURL=`-wrapped inner URLs got percent-mangled / split (breaking `as_canonical_string()` and `translation.py`'s `split("articleURL=")`)
- `?q=a%20b` → `?q=a+b`, valueless `?key` → `?key=`
- signed image/CDN query values re-encoded → broken served image URLs
- every URL with a query string stopped matching its already-stored `url.path` → duplicate `Url`/`Article` rows + broken translated-article cache

Fixed in the latest commit: cleaning is now **surgical** — it splits the raw query, drops only the matched `key=value` segments, leaves survivors byte-for-byte, and returns the original string untouched when nothing matched. Added regression tests for all four cases. 12/12 tests pass.

**Accepted residual:** URL rows stored *before* this change that genuinely had `gaa_*`/`utm_*` in their path won't match the now-cleaned lookup, so a re-encounter creates one fresh canonical row. That's the intended dedup-going-forward behavior; the stale duplicates are harmless. Not worth a backfill.

PR #651, opened 2026-06-05
Daily audio cron: extract decide_generation — table as code

mircealungu:
## Summary
Lifts the per-user generation decision (the table I keep drawing in chat
diagrams) into a single read-only function — one return per row.

```python
def decide_generation(user, language, timezone_offset, today_local):
    sub = DailyAudioSubscription.find(user, language)
    if sub is None or not sub.enabled:                          return ("not-subscribed", None)
    if sub.lesson_type not in VALID_LESSON_TYPES:               return ("invalid-type",  sub)
    if not sub.scheduled_on(today_local):                       return ("not-due",       sub)
    if generator.today_lesson_exists(user, timezone_offset):    return ("exists",        sub)
    if DailyAudioLesson.waiting_paused_for(user, language.id):  return ("paused",        sub)
    return ("ready", sub)
```

The cron's main loop becomes a dispatch on `kind`, and `generate_for_user` does
only the heavy work (it no longer re-checks the gates — `prepare_lesson_generation`
still has its own existing-lesson check as belt-and-suspenders). Both dry-run
and real-run share the same decision, so the buckets can't drift between them.

## Test plan
- [ ] `python -m tools.generate_daily_audio_lessons --dry-run --days 30` — same buckets as before.
- [ ] `--user-id <yours>` real run still generates / skips as before.

Pairs with **web#TBD** (Today view's `selectTodayView` projection — same pattern on the client).

### DIFF
--- a/tools/generate_daily_audio_lessons.py
+++ b/tools/generate_daily_audio_lessons.py
@@ -123,17 +123,29 @@ def resolve_suggestion(user, lesson_type, raw_suggestion):
     return result["canonical"], result["is_general"]
 
 
+def decide_generation(user, language, timezone_offset, today_local):
+    """One row per outcome of "should we generate for this (user, language) today?".
+    Returns (kind, sub) where kind is one of:
+      not-subscribed | invalid-type | not-due | exists | paused | ready
+
+    Read-only: no LLM calls, no writes. Mirrors the daily-audio generation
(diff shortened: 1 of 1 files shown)


mircealungu:
Closing alongside web#1162 (its 'table as code' pair). Daily-audio cron has moved since; re-derive decide_generation on current master if the projection is still wanted.

PR #658, opened 2026-06-30
Updated User.Search to Prioritize exact match and shorter matches

xXPinkmagicXx:
Updated the search function to order by (in priority)

1. Exact match
2. prefix match
3. username length

Also created tests to confirm this behavior in test_user_search.py


mircealungu:
Thanks @xXPinkmagicXx !!!

PR #668, opened 2026-07-05
Added has_leaderboard to cohort

gabortodor:


mircealungu:
ehi mr @gabortodor . i see a draft here. what's the status?

gabortodor:
> ehi mr @gabortodor . i see a draft here. what's the status?

Hi, we have some time to work on the project once again, so hopefully this PR will be mergeable in the following days

klnyzzz33:
@mircealungu we're finished with the has_leaderboards changes, now the teacher can control whether the classroom's students will see the classroom leaderboard or not. See [the frontend PR](https://github.com/zeeguu/web/pull/1184) as well.

We also fixed a small issue that we were aware of with the friend streaks, where it did not increment properly for the very first time.

The last thing remaining that has to be decided, is whether or not the existing users should get their previously earned "hidden" achievements, or start over. Because currently only the newly signed up users get the gamification features, as per our discussion with you:
```python
def _gamification(user: User):
    ...

    if user.id>6479:
        return True

    # Disabled for everyone else
    return False
```

klnyzzz33 (approved):

PR #689, opened 2026-08-01
Friend-share: resolve uploaded-article shares to the promoted original

mircealungu:
## Problem

Sharing a simplification that was made from an **extension/phone upload** stored the wrong `article_id`. [`resolve_original_article_id`](../zeeguu/core/model/shared_article.py) resolves via `parent_article_id` or a `#translated-from-` URL fragment; an upload-based simplification has **neither** (no parent, and a `zeeguu.org/simplified/pending/<uuid>` placeholder URL), so it fell through to `return article.id` — sharing the **A1 copy itself**. The recipient then opens the sharer's adapted copy and can't re-adapt from the original for their own language/level.

## Fix

New `SharedArticle.resolve_shareable_original_id(session, article)`: when there's no parent but the article has a `source_upload`, **promote that upload to a real `Article`** (`Article.find_or_create`, deduped by URL, using the already-captured `text_content` — no network fetch) and share *its* id. The recipient opens the true original (e.g. politiken.dk) and adapts on open, exactly like a crawled share.

Consistent with the shipped "share the original, recipient re-adapts on open" model — no schema change. Chose this (approach A) over a per-user-copy redesign (B) because copies don't actually solve the copyright question either; the real lever is per-article **access control**, deferred as future work.

Safety: if the original won't promote (paywall/parse failure), it falls back to the plain resolver — a share never fails.

## Test plan

- [x] `pytest .../test_friends.py` (24 passed)
- [ ] On device: share an extension/phone-uploaded+simplified article with a friend → their inbox opens the **original** (real publisher), adapts to their level/language — not the sharer's A1 copy.

See [`docs/future-work/track-original-on-send-and-simplify.md`](../docs/future-work/track-original-on-send-and-simplify.md) and [`share-article-to-friend.md`](../docs/future-work/share-article-to-friend.md).

### DIFF
--- a/zeeguu/api/endpoints/friends.py
+++ b/zeeguu/api/endpoints/friends.py
@@ -436,7 +436,7 @@ def share_article_with_friend():
     if not article:
         return make_error(404, "article not found")
 
-    original_id = SharedArticle.resolve_original_article_id(article)
+    original_id = SharedArticle.resolve_shareable_original_id(db.session, article)
     shared = SharedArticle.create(db.session, from_user_id, to_user_id, original_id, note)
 
     # Notify the recipient by email — best-effort, off the request thread.
--- a/zeeguu/core/model/shared_article.py
+++ b/zeeguu/core/model/shared_article.py
@@ -65,6 +65,47 @@ class SharedArticle(db.Model):
         return article.id
 
     @classmethod
+    def resolve_shareable_original_id(cls, session, article) -> int:
+        """Resolve the original article id to store for a share, promoting an
+        upload-based original into a real Article when needed.
+
+        `resolve_original_article_id` handles crawled/translated copies. But a
+        simplification made from a user upload (extension / phone share) has no
+        `parent_article_id` and a placeholder URL, so that resolver would fall
+        through and share the *simplified* copy — leaving the recipient unable to
+        re-adapt from the original. Such articles carry a `source_upload` whose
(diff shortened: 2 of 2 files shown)


mircealungu:
⛔ **Do not merge — superseded by design.**

This resolves the share origin via `find_or_create(upload.url, …)`, which dedupes by URL and returns any existing article as-is. In the common flow (article seen in recommendations = a possibly paywalled/partial crawl → user opens original → sends full body), it grabs the **partial stub** at that URL instead of the full body in the `article_upload`.

Rebuild on the settled design: a `content_provenance` enum + resolve-for-share that attaches to a **full-body** original (promote the upload when needed; never dedupe to a `crawled_partial`). See `docs/future-work/article-body-provenance-and-sharing.md`.

mircealungu:
Superseded. The rebuild lives in the new implementation task, per the settled design in `docs/future-work/article-body-provenance-and-sharing.md` — resolution must prefer the uploaded full body (this PR's `find_or_create`-by-URL grabs a partial crawl instead).

mircealungu:
Superseded by #690 — rebuilt on the corrected model (FK flips onto `article_upload.article_id`; full body stays in the per-user upload, never URL-deduped onto a crawl stub; recipient gets a personalized derivative). See docs/future-work/article-body-provenance-and-sharing.md.

PR #720, opened 2026-08-31
The classroom's rules, in one file — and a table that checks them

mircealungu:
Two commits: a table that **checks** the rules, and a file that **states** them.

## The problem

The rules for what a student sees were spread across `User.cohort_articles_for_user`, `_classroom_only` and the article filters. No file stated any of them. That is the common root of the last fortnight's bugs — the Greek learner staring at a blank app, the 54 students with texts in their class they cannot open, the teacher exemption that made the first bug report misleading.

## 1. `zeeguu/core/classroom.py` — the rules, readable

The whole of the classroom's behaviour in the order someone would ask about it. Each rule is one function and the sentence that justifies it:

```python
def student_can_read(student, text):
    """A student reads only in the language they are currently learning.

    The student's own setting decides, not the class's -- which is why a
    Greek learner in an English class sees an empty classroom, and why a class
    can hold texts that nobody in it can open.
    """

def classroom_of(student):
    """The texts a student sees, each with the classes it reached them through.

    One merged list: a student in several classes has one classroom, not one
    per class. A text shared with two of their classes appears once, naming both.
    """

def sees_only_class_texts(user):
    """Whether the app is reduced to the classroom: no feed, no search, no inbox.

    ... the strictest class wins ... Teachers are never restricted, even in a
    class that asks for it: they need the feed and the search to find the texts
    they are going to share.
    """
```

Both former homes become adapters: `cohort_articles_for_user` turns the answer into article infos and tags them; `_classroom_only` forwards.

## 2. `test_classroom_visibility_rules.py` — the same rules as a table

```
#        learning  classes                                sees  mode
("a class in another language is invisible, not empty",
     "da",     {"English Class": "en en"},              0,   FULL_APP),

("a class may hold texts the student cannot see",
     "da",     {"Mixed Class": "da en en"},             1,   FULL_APP),

("Jack's bug: restricted, and every text is in another language",
     "de",     {"CUT!": "en en en en en"},              0,   CLASSROOM_ONLY),
```

A row is one world. `!` on a class name means "students see only the texts I share". Each row builds that world and asks the real code.

## No behaviour change, and the table is what says so

**298 tests and 20 subtests pass unchanged.** The refactor is exactly the kind that is nerve-racking without a spec and dull with one.

The table also found a live fragility on its first run: `_tiago_exercises` dereferenced `user.learned_language.code` with no guard, and `features_for_user` runs on every `/user_details` — an account with no learned language would have taken down the whole call. Zero such accounts exist today; `cohort_articles_for_user` already handled the case, so the two halves disagreed. Guarded.


mircealungu:
Closing in favour of two smaller PRs. The refactor moved two one-line rules into a three-file package without making the code simpler (net +76 production lines) and left the real duplication untouched. What's kept: the test table and the _tiago_exercises guard → #759; the duplicated 'uploaded texts only through a current class' filter (recommender + DB fallback) now has one home → #758.

